PR #17, opened 2020-07-03
mention n8n-heroku as alternative deploy option

einSelbst:
I used n8n-heroku successfully.


tanay1337:
Hi @einSelbst, thanks for sending in the PR. Unfortunately, we can't merge this PR because the link that you have added is an external GitHub project, which is not maintained by n8n. Hence, we cannot include this in the official documentation. 

Nonetheless, thanks for taking out the time to think about this :)

PR #18, opened 2020-07-05
Add link from dev docs to create-nodes docs

einSelbst:
I sometimes had trouble finding the correct docs as are multiple pages on the topic. Maybe even more links from one doc to another could be helpful.

https://github.com/n8n-io/n8n/blob/master/CONTRIBUTING.md
https://github.com/n8n-io/n8n/tree/master/packages/node-dev

https://docs.n8n.io/reference/development.html
https://docs.n8n.io/nodes/creating-nodes/create-node.html


einSelbst:
I think there is still a page missing where I did read about how to extend existing nodes. I will reopen the PR when I have found that page. Sorry for the noise.

PR #28, opened 2020-07-08
Fix GitHub casing

ivov:
This fixes the casing of `Github` to `GitHub` throughout the docs.

This affects display instances of the word, not functional instances like URLs or permalinks.

The single non-display fix is the new dir name `/docs/nodes/credentials/GitHub`


tanay1337:
Merged, thanks a lot @ivov 🎉

PR #33, opened 2020-07-10
Google Sheets docs

ivov:
Here are the Google Sheets docs!

The new example workflow is [here](https://n8n.io/workflows/449). I also added OAuth2 credentials instructions.

Let me know if you have any comments!


tanay1337:
Merged, thanks for the great work, @ivov 🎉

PR #101, opened 2020-08-11
Minor reordering for QuestDB docs

smamudhan:


tanay1337:
@smamudhan If you follow the suggested change in the workflow, the workflow would fail to run. Can you confirm if that works for you or if I am missing something here?

tanay1337:
Based on our discussion @smamudhan, closing this PR.

PR #102, opened 2020-08-11
Enhanced CrateDB docs

smamudhan:


tanay1337:
@smamudhan The workflow here is incorrect and does not insert data correctly.

image

PR #178, opened 2020-09-15
Changelog update for n8n@0.82.0

harshil1712:


tanay1337 (changes requested):
Shared feedback on the call.

PR #321, opened 2020-12-08
Fix link to workflow creation overview section

ivov:
At the [Cron page](https://docs.n8n.io/nodes/n8n-nodes-base.cron/), the link [Creating Your First Workflow](https://docs.n8n.io/getting-started/creating-your-first-workflow.html) gives a 404.


tanay1337:
Thanks a lot @ivov, nice catch :)

PR #377, opened 2021-01-19
Update README.md of FunctionItem & function-nodes.md

ixidion:
I fully understood the FuntionItem node in the end from checking the source code, but the documentation was confusing for me in several ways. That's because Function and FunctionItem have different styles. Function gives access to the whole items array and the structure is transparently available with json/binary.
In case of FunctionItem. The item is not item, it's item.json. You can't access item.binary cause of this, as it would be intuitive to do so, if you worked with Function before.
At first an update to the documentation to guide new users hopefully in the right direction. 

It should be also considered that Function and FunctionItem behave the same, e.g. item.json and item.binary should exists. The variable might also called itemjson instead of item.


tanay1337:
Merged, thanks a lot @ixidion :D

PR #408, opened 2021-02-08
Doc | Function | Add reference to input contract

k-metric:
Added a line referencing the contract data structure between nodes because Function seems a bit low level compared to other nodes. Please feel free to edit. Thanks.


tanay1337:
Thanks a lot for contributing @k-metric, got merged!

PR #421, opened 2021-02-17
Update docker-compose to fix LetsEncrypt config

janbaykara:
This PR updates [the docker tutorial](https://docs.n8n.io/reference/server-setup.html#_5-create-docker-compose-file). Expose port 80 so LetsEncrypt doesn't fail, but redirect it to port 443.

It looks like the current docker-compose setup doesn't correctly enable LetsEncrypt for multiple users. [There's a forum thread about this.](https://community.n8n.io/t/ssl-and-security/346/16?u=janbaykara)

### DIFF
--- a/docs/reference/server-setup.md
+++ b/docs/reference/server-setup.md
@@ -95,11 +95,15 @@ services:
       - "--api.insecure=true"
       - "--providers.docker=true"
       - "--providers.docker.exposedbydefault=false"
+      - "--entrypoints.web.address=:80"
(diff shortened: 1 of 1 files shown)


janober:
Link community discussion for reference:
https://community.n8n.io/t/ssl-and-security/346

gpene:
Added as part of https://github.com/n8n-io/n8n-docs/commit/217836e51cb750774e7a17f704bea44a0694a0e6 - thank you @janbaykara

PR #423, opened 2021-02-18
Listing all files/folders within a folder

arjones:
It took me a while figure out how to use Query String to filter the results from Google Drive Node.
I added a FAQs session where I wrote about this configuration, I think it can be useful to more users.


tanay1337:
Merged here: https://github.com/n8n-io/n8n-docs/pull/429. Thanks a lot for contributing @arjones :D

PR #430, opened 2021-03-02
Add example for format

harshil1712:


docs/nodes/nodes-library/core-nodes/DateTime/README.md
```diff
@@ -53,7 +53,22 @@ The start node exists by default when you create a new workflow.
...
+You can use the following values in the ***Form Format*** field:
+
+| Input | Example | Description |
+|-------|---------|-------------|
+|X | 1616108400 | Unix timestamp |
+|x | 1616108400000 | Unix ms timestamp |
+| DD-MM-YYYY | 19-03-2021 | Date, month, and year separated by `-` |
+| DD/MM/YYYY | 19/03/2021 | Date, month, and year separated by `/` |
+
+You can also use the options available in the ***To Format*** dropdown list.
```
tanay1337, line 74: Since we are doing a reference table, we should then really add all the possible formats in there.

docs/nodes/nodes-library/core-nodes/DateTime/README.md
```diff
@@ -30,7 +30,7 @@ You can specify the format that the date has to be converted to from the *To For
 You can also specify a custom format by setting the *Custom Format* toggle to 'On'.
 
 - Options
-	- *From Format* field: Allows you to specify the format of the input values.
+	- *From Format* field: Allows you to specify the format of the input values. For example, `YYYY/MM/DD HH:MM:SS`, `DD.MM.YYYY`, `MM/DD/YYYY`, etc. Refer to the FAQs to learn about the possible vaules.
```
tanay1337, line 33: We should link to the FAQs below.

tanay1337 (changes requested):

PR #437, opened 2021-03-08
Fixed wrong version number in commit comparison

phizab:


tanay1337:
Thanks a lot @phizab, would you mind signing the CLA so that I can merge this? :)

PR #469, opened 2021-04-07
Add FAQs for data type functionality

harshil1712:


docs/nodes/nodes-library/nodes/CrateDB/README.md
```diff
@@ -63,6 +63,11 @@ The start node exists by default when you create a new workflow.
 
 ![Using the CrateDB node to insert data into a table](./CrateDB1_node.png)
 
+## FAQs
+
+### How to specify the data type of a field?
```
tanay1337, line 68: @harshil1712 Which field are you referring to here? Does that work for all fields, e.g., table, and so on?

docs/nodes/nodes-library/nodes/CrateDB/README.md
```diff
@@ -63,6 +63,11 @@ The start node exists by default when you create a new workflow.
...
+To specify the data type of a field, append the field name with `:type`, where `type` is the data type of that field. For example, if you want to specify the type `int` for the field `id` and type `text` for the field `name`, you can use the following snippet `id:init,name:text`.
```
tanay1337, line 69: Does it work for all the databases? I know that not all the DBs support the same actions because of the underlying differences.

tanay1337 (changes requested):

docs/nodes/nodes-library/nodes/CrateDB/README.md
harshil1712: According to [this comment](https://github.com/n8n-io/n8n/pull/1600#issue-602804836) it was tested and works for all the DBs.

PR #492, opened 2021-04-19
Updating OpenWeatherMap API Instructions

zromick:
Added a note about OpenWeatherMap's API key activation being 10 min to 2 hours after a successful registration. 
Users could become frustrated after following the steps and getting an error (including me, which is why I sought to update the docs).

See: https://openweathermap.org/faq#notes_block:~:text=Do%20I%20need%20to%20activate%20my%20API%20key%3F


zromick:
Wording might be better as "at least 10 minutes after successful registration" to prevent people from disengaging with the content.

tanay1337:
Many thanks for submitting this @zromick. This will be very helpful for loads of folks! :D

PR #498, opened 2021-04-25
Update docker-compose to pass timezone env variable

SimonMayerhofer:
GENERIC_TIMEZONE is added in the sample .env file but not passed to the docker container through the sample docker-compose.yml


tanay1337:
Thanks a ton for the PR @SimonMayerhofer. You rock! 🙌

PR #503, opened 2021-04-27
Add spring cleaning redirects template

ivov:
This draft PR is a proposal for adding the spring cleaning redirects with the [`addRoute`](https://router.vuejs.org/api/#router-addroute) method.

Notes: 
- The `addRoutes` method (plural) is deprecated.
- Adjustments will be needed if new sections are created.

/cc @harshil1712


harshil1712:
@ivov I tested this out and it works wonderfully!

PR #545, opened 2021-06-09
Fix Typo

pReya:
* fix typo
* add proper link to the mentioned section (which seems to have moved and is not "below" anymore)


tanay1337:
Thanks a lot for the PR, @pReya 🙌

PR #546, opened 2021-06-09
Reimplement data fetching, add icons for credentials

mutdmour:
- load nodes and credentials directly from nodes-base during compile time, instead of a graphql request client side... now all pages are static again
- use fontawesome for core node icons
- add icons for credentials page, using any relevant node that uses one


mutdmour (commented):
hints

harshil1712:
Thank you @mutdmour! All the issues have been addressed in this PR 💖

PR #549, opened 2021-06-16
Add feedback plugin

harshil1712:


ivov:
Added a layout to embed `<Feedback />` in all pages.

Hopefully this is what you had in mind - otherwise feel free to revert.

PR #552, opened 2021-06-18
:sparkles: Add docs for SSH node

gpene:


docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
...
+permalink: /nodes/n8n-nodes-base.ssh
+description: Learn how to use the SSH node in n8n
+---
+
+# SSH
+
+The SSH node is useful for executing commands via the Secure Shell Protocol.
+
+::: tip 🔑 Credential
+Authentication for the SSH node requires that you have a username and password for connecting to the remote server, or an SSH key configured for the server or service you are connecting to. See [Connecting to GitHub with SSH] for an example](https://docs.github.com/en/github/authenticating-to-github/connecting-to-github-with-ssh).
```
harshil1712, line 11: This doesn't render correctly. Can we fix the markdown issue for the Link?

harshil1712: Can we also add separate documentation for the credentials? Example: https://docs.n8n.io/credentials/smtp/#send-email

docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
...
+- ***Command:*** Only visible for **Command** resources. The command to execute on the remote machine.
+- ***Working Directory:*** Only visible for **Command** resources. The directory where the command should be executed.
+- ***Path:*** Only visible for **Download** operation on file resources. The path where the desired file is found.
+- ***Binary Property:*** Only visible for file resources. The name of the binary property which contains the data for the file to be uploaded.
+- ***Target Directory:*** Only visible for **Upload** operations on file resources. The directory to upload the file to. The name of the file does not need to be specified, it's taken from the binary data file name. To override this behavior, set the parameter **File Name** under options.
+
+
+## Example Usage
+
+This workflow allows you to remotely connect to your computer via SSH. This example usage workflow would use the following nodes.
```
harshil1712, line 36: Can we have a more useful example? The example in its current state isn't helpful 🙂

harshil1712: Also, please share the workflow on the workflow page and share the link here. This helps people to try out the workflow if they want to create it from scratch

docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+![A workflow with the SSH node](./workflow.png)
+
+### 1. Start node
+
+The start node exists by default when you create a new workflow.
+
+### 2. SSH node
+
+1. From the SSH Credentials dropdown, select ***Create New*** and complete the following fields:
```
harshil1712, line 50: This can go in the credentials documentation.

docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
...
+    - ***User***: Enter the your username for this server.
+    - ***Password***: Enter your password for the provided username.
+    - ***Private Key***: Only available for Private Key authentication. Enter your SSH key.
+    - ***Passphrase***: Only available for Private Key authentication. Enter your passphrase.
+
+![SSH credentials](./ssh_credentials.png)
+
+2. Select ***Password*** from the ***Authentication*** dropdown list.
+3. Select ***Command*** from the ***Resource*** dropdown list.
+4. The ***Execute*** option is automatically selected as the ***Operation***.
```
harshil1712, line 63: We mention something like this if they have to select a different operation. But I like this as well. Instead of using the word automatically, can we use the word default?

docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
...
+5. In the ***Command*** field enter the command to execute. Here we used `ls` just to see the local directory contents.
+6. In the ***Working Directory*** field you can enter where the command should be executed. Here we ran it in the root directory.
+7. Click on ***Execute Node*** to run the node.
+
+![Executed Workflow](./executed_ssh_workflow.png)
```
harshil1712, line 68: The Alt text and the image name don't follow the standard we use.

docs/nodes/nodes-library/core-nodes/SSH/README.md
```diff
@@ -0,0 +1,68 @@
+---
+permalink: /nodes/n8n-nodes-base.ssh
+description: Learn how to use the SSH node in n8n
+---
+
+# SSH
+
+The SSH node is useful for executing commands via the Secure Shell Protocol.
```
harshil1712, line 8: Do you think it might be a good idea to give a few cases as an example here? This will help non-technical users know how they can use this node.

harshil1712 (changes requested):

docs/nodes/nodes-library/core-nodes/SSH/README.md
gpene: Done 👍  Though in general I disagree with the practice of adding a credential page regardless of the node - the Send Email one being a perfect example. That page adds no value - the same could be accomplished in one line in the node docs, and without forcing the user to navigate away to a separate page.

docs/nodes/nodes-library/core-nodes/SSH/README.md
gpene: If they want to create it from scratch then sharing the workflow for them to copy defeats that purpose. Also, unless it is a particularly complex and useful example, I do not think each and every workflow should be uploaded to the Workflows page - it creates a mess, making it more difficult to find useful things on that page, without adding much value.

docs/nodes/nodes-library/core-nodes/SSH/README.md
gpene: I don't believe there is any standard for this mentioned in the Contributing page - if it will be enforced that it should be included there.

docs/nodes/nodes-library/core-nodes/SSH/README.md
gpene: Agree that a more useful example is needed. I have asked if there are any resources available internally I could use.

harshil1712: I've updated my previous comment. We share the workflows on the workflow page for people who want to try out the example without creating the whole workflow.

docs/nodes/nodes-library/core-nodes/SSH/README.md
harshil1712: Ah yes. we don't have guidelines for this yet. But if you refer to other node documentation, you will get an idea of what I mean. The alt text is really helpful for people who use a screen reader.

docs/nodes/nodes-library/core-nodes/SSH/README.md
harshil1712: This helps people who are not technical. If you check the IMAP node, we added docs for different services. In the future, if the community is facing issues with creating credentials and connecting the SSH node, this is where the documentation would go 🙂

PR #585, opened 2021-07-23
Fix ServiceNow credentials doc link

harshil1712:
This PR fixes the link for the ServiceNow credential documentation. The node in n8n refers to the link with a capitalized N (serviceNow) and that results in a 404.

Screenshot 2021-07-23 at 11 31 08 AM


gpene:
Fixed separately - thanks Harshil

PR #586, opened 2021-07-24
Fix HomeAsssistant credential ReadMe

kylefmohr:
Fix incorrect link in Home Assistant credential readme. It was linking to Harvest.


gpene:
Thank you @kylefmohr 🙏

PR #592, opened 2021-08-05
Update create-trigger-node.md, should be `onReceived` not `OnReceived`.

chasewoo:
should be `onReceived` not `OnReceived`.


gpene:
Thanks @chasewoo!

PR #615, opened 2021-08-30
Update README.md - Notion Trigger Node instruction small correction.

SalvatoryMalamar:
The Notion Trigger Node instructions reference a non-existent event type. "Record Added" no longer appears to exist. The only event type available is "Page Added to Database." So, a proposal to update the Notion Trigger Node instructions. Corrected the event type from "Record Added" to "Page Added to Database"


gpene:
Thank you @SalvatoryMalamar - merged 👍

PR #620, opened 2021-09-02
Uses code block to avoid emoji

kinduff:
**Before**

Screenshot from 2021-09-02 11-51-32

**After**

Screenshot from 2021-09-02 11-54-30


gpene:
Thank you @kinduff !

PR #625, opened 2021-09-15
Update README.md

jaikanth6:
I spent a good deal of time trying to understand why the error trigger node wasn't working and after a lot of poking around in the forums, found this message from Harshil that it works / triggers only when the workflows are executed automatically. Not sure why this critical information is missing in the documentation. I can see a lot of questions in the forums regarding the "error trigger node" not working.


gpene:
Thanks @jaikanth6 , the FAQ at the bottom of the page does state it can't be triggered manually but this makes it clearer and more visible. Thank you 🙏

PR #630, opened 2021-09-23
Parse versioned nodes

ivov:
Mattermost is the first and for now only versioned node - the plugin cannot parse it, which breaks the build.

To test the fix, uncomment [here](https://github.com/n8n-io/n8n-docs/commit/9aa3c868c6e6c6a271d3b245374b3f16fdb7612d).

### DIFF
--- a/docs/.vuepress/plugins/nodes.js
+++ b/docs/.vuepress/plugins/nodes.js
@@ -40,7 +40,14 @@ async function getNodes() {
 
 	const descriptions = {};
 	for (let name in nodes) {
-		const node = nodes[name].description;
+		let node;
+
+		if (nodes[name].nodeVersions) {
+			const maxVersion = Math.max(...Object.keys(nodes[name].nodeVersions));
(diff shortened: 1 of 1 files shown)


docs/.vuepress/plugins/nodes.js
```diff
@@ -40,7 +40,14 @@ async function getNodes() {
 
 	const descriptions = {};
 	for (let name in nodes) {
-		const node = nodes[name].description;
+		let node;
+
+		if (nodes[name].nodeVersions) {
+			const maxVersion = Math.max(...Object.keys(nodes[name].nodeVersions));
```
mutdmour, line 46: for documentation reasons, should not we be supporting versions? if operations changed between versions, how do we point users to previous version?

ivov: We should definitely document versions properly at some point. This is a hotfix to unlock the build.

mutdmour (approved):

PR #636, opened 2021-10-05
Fix inconsistency on data flow section

mrchrisadams:
When I read this section about data flow with the trello example,  it looked like the data flow example should be mentioning cards not boards for Trello.


gpene:
Thanks @mrchrisadams

PR #646, opened 2021-10-29
add the dropcontact documentation

PaulineDropcontact:
with the new dropcontact node

with love 💚

PR #654, opened 2021-11-08
Revert "add the dropcontact documentation"

that-one-tom:
Reverts n8n-io/n8n-docs#646


that-one-tom:
Reverting this PR as the build fails

PR #655, opened 2021-11-08
Fix Dropcontact documentation

harshil1712:
Hey @PaulineDropcontact,

We had to roll back the documentation. Here's quick feedback:
- [ ] Add the example workflow to the [workflow page](https://n8n.io/workflows) and add a link in the docs. (hint: I've added a comment where you can update the URL)
- [ ] Update screenshots
- [ ] I tried following the steps you mentioned in the example workflow, but it seems to miss some steps. Can you please add all the steps? You can look at the [Google Sheets node documentation](https://docs.n8n.io/nodes/n8n-nodes-base.googleSheets) for inspiration.

Let me know if you need any help :)


harshil1712:
Updates from #656 merged.

PR #689, opened 2021-12-17
Update desktop users

djangelic:
Currently customers are being sent to this page to update their desktop versions but there are no instructions on this page to let the customer know that an update is not currently avaiable. I've added a new header for N8n desktop along with a short blurb explaining that it is not updatable yet but should be soon.


harshil1712:
Hey @djangelic,

Thank you for opening this PR! We now have a new desktop version and we have updated the instructions in #692!

djangelic:
Thank you! It looks great!

PR #727, opened 2022-02-13
Update README.md

ovann86:
Change "sensible" to "sensitive". Appears the wrong word was used, better explains why you want self hosted.


StarfallProjects:
Good spot - thank you.

PR #728, opened 2022-02-13
Update README.md

meyer-roessl:
Filemaker is not open source. Also, the path has changed (filemaker.com redirects to the new path).


StarfallProjects:
Thanks for the fix!

PR #744, opened 2022-02-23
Workflow templates

StarfallProjects:


StarfallProjects:
Do not merge until https://github.com/n8n-io/n8n/pull/2720 is done

docs/getting-started/key-components/workflow.md
```diff
@@ -55,4 +55,40 @@ If your workflow execution fails, you can retry the execution. To retry a failed
...
+### Use a workflow template
+
+1. In the sidebar, click **Templates**.
+2. Browse or search the workflow templates list.
+3. Click a workflow to view more information. n8n opens the workflow details page.
+4. On the workflow details page, click **Use this workflow**. n8n opens the workflow.
+5. Click **Save** to add the workflow to your workflows.
+
+::: tip 
+Workflow templates are available on all flavours of n8n. If you can't access workflow templates in n8n, you may be on an old version, or using a self-hosted or embeded version of n8n with templates disabled.
```
lorenanda, line 78:
```suggestion
Workflow templates are available on all flavors of n8n. If you can't access workflow templates in n8n, you may be on an old version, or using a self-hosted or embedded version of n8n with templates disabled.
```

It would be more clear to mention the first version on which templates are available, for example: "you may be on an old version (< 0.xxx.x)".

docs/getting-started/key-components/workflow.md
```diff
@@ -55,4 +55,40 @@ If your workflow execution fails, you can retry the execution. To retry a failed
...
+1. In n8n, download your workflow JSON: 
+    1. Open the workflow
+    2. Click **Workflows** > **Download**.
+2. [Log in](https://n8n.io/login) to your n8n dashboard.
+3. Click **Share New Workflow**.
+4. Enter your workflow details:
+    * **Name**: this should be descriptive but simple. It will appear on the website and in app.
+    * **Description**: tell users what the workflow does. Include any configuration or setup steps.
+    * **Workflow code**: copy in the workflow JSON that you downloaded.
+5. Click **Publish Workflow to Share**. n8n reviews all workflows before publishing them.
```
lorenanda, line 94:
```suggestion
5. Click **Publish Workflow to Share**.

Note that n8n reviews all workflows before publishing them.
```
I'd add this information on a new line (maybe even make it a callout) because it's related to the entire template sharing process. Also (if we have this info), I'd add a brief explanation on how n8n's reviewing might impact the user's publishing (e.g. it might take a while for the template to appear in the library, or it might be rejected if it doesn't follow some requirements).

docs/getting-started/key-components/workflow.md
```diff
@@ -55,4 +55,40 @@ If your workflow execution fails, you can retry the execution. To retry a failed
...
+You can also use the [Error Trigger node](../../nodes/nodes-library/core-nodes/ErrorTrigger), which triggers a workflow when another workflow has an error. Once a workflow fails, this node gets details about the failed workflow and the errors.
+
+## Workflow templates
+
+When creating a new workflow, you can choose whether to start with an empty workflow, or use an existing template.
+
+Templates provide:
+
+* A way to get started quickly: we might already have a template that does what you need.
+* Examples of what you can build, and best practices for creating your own workflows.
```
lorenanda, line 67:
```suggestion
* Examples of what you can build
* Best practices for creating your own workflows
```

I just see them as two separate benefits.

lorenanda (changes requested):
Looks good to me, only made some minor suggestions.

docs/getting-started/key-components/workflow.md
StarfallProjects: @lorenanda this one I'm leaving as-is, precisely because we don't have enough info - I asked, and we basically don't have a clear review process or timescales at the moment. I'd very much like to add that info once we have it.

docs/getting-started/key-components/workflow.md
StarfallProjects: Will add the version once we know it :-)

PR #752, opened 2022-03-10
Doc152 user management

StarfallProjects:


StarfallProjects:
Merge at the same time as user management goes live.

docs/reference/environment-variables.md
```diff
@@ -82,6 +82,13 @@ Enabling overwrites for credentials allows you to set default values for credent
 | `N8N_DEFAULT_BINARY_DATA_MODE` | `string` | The default binary data mode. Default value is `default` and keeps binary data in memory. Set to `filesystem` in order to use the filesystem. |
 | `N8N_PERSISTED_BINARY_DATA_TTL` | `number` | Time to live (in minutes) for persisted data. Default is `1440`. |
 | `VUE_APP_URL_BASE_API` | `string` | Used when building the `n8n-editor-ui` package manually to indicate how the frontend can reach the backend API. |
+| `N8N_USER_MANAGEMENT_DISABLED` | Boolean | Defaults to user management enabled (false). Set to `true` to disable the [user management](user-management.md) feature. |
```
sirdavidoff, line 85: This only works if you haven't yet set up an owner account

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+
+The **Users** page shows all users, including ones with pending invitations. You can:
+
+* Delete a user: 
+  1. Click the menu icon by the user you want to delete.
+  2. Confirm you want to delete them.
+  3. If they are an active user, choose whether to copy their workflow data and credentials to a new user, or permanently delete their workflows and credentials.
+* Resend an invitation to a pending user: click the menu icon by the user, then click **Resend invite**.
+
+## Account types
```
sirdavidoff, line 73: Feels like this should go at the top

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+### Step one: SMTP
+
+You need an SMTP server for user management to send invites and password resets. Get the following information from your SMTP provider:
+
+* Server name
+* SMTP username
+* SMTP password
+* SMTP sender name
+
+To set up SMTP with n8n, configure the following environment variables for your n8n instance:
```
sirdavidoff, line 28: Should have a link here to something that explains how to apply environment vars

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
+# User management
+
+User management in n8n allows you to invite people to work in your n8n instance. It includes:
+
+* Login and password management
+* Adding and removing users
+* Two account types: owner and member
+
+User management is available for self-hosted n8n. It isn't currently available for Cloud.
```
sirdavidoff, line 9: Not on Desktop either

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+
+* Leave it enabled, but choose to skip the setup step. You can use n8n as normal. If you want to set up user management later, go to **Settings** > **Users**.
+* Disable the feature completely using the `N8N_USER_MANAGEMENT_DISABLED` environment variable. Setting this environment variable to `true` completely hides the feature in your n8n instance.
+
+## Best practices
+
+This sections contains advice on best practices relating to user management in n8n.
+
+* We recommend that owners create a member-level account for themselves. Owners can see all workflows, but there is no way to see who created a particular workflow, so there is a risk of overriding other people's work if you build and edit workflows as an owner.
+* Multiple users can't edit the same workflow simultaneously. n8n does not support this type of synchronous collaboration.
```
sirdavidoff, line 104: You can edit simultaneously. You'll just overwrite each others' changes

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+* Webhook paths must be unique across the entire instance. This means each webhook path must be unique for all workflows and all users. By default, n8n generates a long random value for the webhook path, but users can edit this to their own custom path. If two users set the same path value:
+    * The path works for the first workflow that registers it (n8n registers workflow paths the first time the workflow activates or runs).
```
sirdavidoff, line 106: Can we simplify this to 'The path works for the first workflow that is run or activated'?

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+    * Other workflows will error if they try to run with the same path.
```
sirdavidoff, line 107: How does the user get out of this situation?

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+::: tip Create a member-level account for the owner
+We recommend that owners create a member-level account for themselves. Owners can see all workflows, but there is no way to see who created a particular workflow, so there is a risk of overriding other people's work if you build and edit workflows as an owner.
+:::
+
+## Skipping or disabling user management
+
+You don't have to use n8n's user management feature. You can:
+
+* Leave it enabled, but choose to skip the setup step. You can use n8n as normal. If you want to set up user management later, go to **Settings** > **Users**.
+* Disable the feature completely using the `N8N_USER_MANAGEMENT_DISABLED` environment variable. Setting this environment variable to `true` completely hides the feature in your n8n instance.
```
sirdavidoff, line 97: Only if you haven't set up a member account

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
...
+| -------- | ---- | ----- |
+| `N8N_SMTP_EMAIL_MODE` | string | smtp |
+| `N8N_SMTP_HOST` | string | _your_server_name_ |
+| `N8N_SMTP_USER` | string | _your_smtp_username_ |
+| `N8N_SMTP_PASSWORD` | string | _your_smtp_password_ |
+| `N8N_SMTP_SENDER` | string | _your_smtp_sender_name_ |
+
+If your n8n instance is already running, you need to restart it to enable the new SMTP settings.
+
+::: tip New to SMTP?
```
sirdavidoff, line 40: What if this went first in this section?

docs/reference/user-management.md
```diff
@@ -0,0 +1,107 @@
+# User management
```
sirdavidoff, line 1: FYI there is new info at the bottom of this doc that could go under 'best practices': https://www.notion.so/n8n/P0-UM-Docs-notes-ec67f85a5f0d46c08c58ff0862926ba3

Plus it seems like not everything mentioned in the alpha testing doc is in here?

sirdavidoff (changes requested):
See inline comments

docs/reference/environment-variables.md
StarfallProjects: Added note

docs/reference/user-management.md
StarfallProjects: Moved

docs/reference/user-management.md
StarfallProjects: Added

docs/reference/user-management.md
StarfallProjects: Changed to:

Users must be careful not to edit the same workflow simultaneously. It is possible to do it, but the users will overwrite each other's changes.

docs/reference/user-management.md
StarfallProjects: Added

docs/reference/user-management.md
StarfallProjects: Changed (was trying to avoid passive voice, but perhaps trying too hard)

docs/reference/user-management.md
StarfallProjects: @sirdavidoff  A member account or an owner account?

docs/reference/user-management.md
StarfallProjects: The things missing from the alpha testing doc are the features you can turn off if you don't want them (tags, templates, survey). However, these are not specific to user management. I've added another tip linking to the environment variables list and calling out those options.

docs/reference/user-management.md
StarfallProjects: I'd prefer to leave it where it is. Get everyone to the key info first, then provide more for those that need it. Almost the same principle as headline -> summary -> full detail.

docs/reference/environment-variables.md
```diff
@@ -82,6 +82,13 @@ Enabling overwrites for credentials allows you to set default values for credent
 | `N8N_DEFAULT_BINARY_DATA_MODE` | `string` | The default binary data mode. Default value is `default` and keeps binary data in memory. Set to `filesystem` in order to use the filesystem. |
 | `N8N_PERSISTED_BINARY_DATA_TTL` | `number` | Time to live (in minutes) for persisted data. Default is `1440`. |
 | `VUE_APP_URL_BASE_API` | `string` | Used when building the `n8n-editor-ui` package manually to indicate how the frontend can reach the backend API. |
+| `N8N_USER_MANAGEMENT_DISABLED` | Boolean | Defaults to user management enabled (false). Set to `true` to disable the [user management](user-management.md) feature. Note that this doesn't work if you have already set up an owner account. |
```
sirdavidoff, line 85: Not 100% clear to me what 'doesn't work'. Perhaps "Note this flag is ignored if you've already...'?

sirdavidoff (changes requested):
Almost there! Left a comment or two.

The only other thing I would suggest: I don't know if tags are important enough to have at the top under what each type of user can do. I wonder if it would be better to keep them as a tip at the bottom. But your call — I can see how they logically fit where you've put them.

(No need to review again after this BTW)

docs/reference/user-management.md
```diff
@@ -0,0 +1,118 @@
...
+
+* We recommend that owners create a member-level account for themselves. Owners can see all workflows, but there is no way to see who created a particular workflow, so there is a risk of overriding other people's work if you build and edit workflows as an owner.
+* Users must be careful not to edit the same workflow simultaneously. It is possible to do it, but the users will overwrite each other's changes.
+* Webhook paths must be unique across the entire instance. This means each webhook path must be unique for all workflows and all users. By default, n8n generates a long random value for the webhook path, but users can edit this to their own custom path. If two users set the same path value:
+    * The path works for the first workflow that is run or activated.
+    * Other workflows will error if they try to run with the same path.
+* If you run n8n behind a reverse proxy, set the following environment variables so that emails are generated with the correct URL:
+  * `N8N_HOST`
+  * `N8N_PORT`
+  * `N8N_PROTOCOL`
```
csuermann, line 116: This should be updated to reflect the new `N8N_EDITOR_BASE_URL` environment variable. 
(cf. [commit](https://github.com/n8n-io/n8n/pull/2967/commits/adc0f32c52784e77a42262ffcb39c776fda7bdb6))

docs/reference/environment-variables.md
StarfallProjects: Changed

docs/reference/user-management.md
StarfallProjects: Added

docs/reference/user-management.md
sirdavidoff: Sorry, owner account

StarfallProjects: Done

PR #753, opened 2022-03-10
📚 Added 0.166.0 changelog

that-one-tom:


docs/reference/changelog.md
```diff
@@ -2,6 +2,39 @@
...
+
+### New nodes ✨
+
+<br />
+<Changelog node="n8n-nodes-base.odoo" title="Odoo" />
+
+### Enhanced nodes 🚀
+
+<br />
+<Changelog node="n8n-nodes-base.function" title="Function:" text="Added support for items without a json key."/>
```
StarfallProjects, line 18:
```suggestion
<Changelog node="n8n-nodes-base.function" title="Function:" text="Added support for items without a JSON key."/>
```

docs/reference/changelog.md
```diff
@@ -2,6 +2,39 @@
...
+
+### Core functionality ⚙️
+
+- Added new environment variable `N8N_HIRING_BANNER_ENABLED` to enable/disable the hiring banner.
+- Fixed a bug preventing keyboard shortcuts from working as expected.
```
StarfallProjects, line 23: Should these be under bug fixes?

StarfallProjects (changes requested):

docs/reference/changelog.md
that-one-tom: Afaik the bug fix section was only ever for node-specific bug fixes using the `<Changelog>` tag

StarfallProjects: Ah ok. We probably ought to retitle the bugfixes section to "Node bugfixes" then? As it looks weird to have a load of bugfixes, followed by a bugfixes section

PR #779, opened 2022-03-23
updated NAT IP

StarfallProjects:


docs/reference/faq.md
```diff
@@ -133,5 +133,5 @@ We're always excited to talk about software licenses, fair-code and other princi
 
 ### What is the IP address of n8n.cloud?
 
-Currently it is `20.79.72.105`, however this is subject to change. Also the NAT address is `20.79.72.36`, but this is also subject to change.
-Recommended practice is to whitelist `20.79.72.0/24`, but if more strict measures are needed, at minimum both `20.79.72.105` and `20.79.72.36` must be whitelisted.
+Currently it is `20.79.72.105`, however this is subject to change. Also the NAT address is `20.79.227.226`, but this is also subject to change.
```
that-one-tom, line 136: It looks there are now two NAT addresses, `20.79.227.226` and `20.79.72.36`. So maybe go with something like "The NAT addresses are `20.79.227.226` and `20.79.72.36`..." over "Also the NAT address is `20.79.227.226`"?

that-one-tom (changes requested):

PR #785, opened 2022-03-28
Add a hint for providing multiple recipients

vuchl:


StarfallProjects:
@vuchl thanks for this. I took the liberty of moving the note up to be with the rest of the email recipient info. Really helpful addition!
It looks like you may not have signed the CLA - you need to do this before we can merge.

PR #844, opened 2022-05-19
Environment variables refresh

StarfallProjects:


docs/hosting/configuration.md
```diff
@@ -117,11 +116,11 @@ The following environment variables support file input:
 
 ### Base URL
 
-!!! warning "Keep in mind"
-    This variable only gets used when the `n8n-editor-ui` package gets built manually. Hence, it does not get used in combination with the default n8n docker image. By default, `/` gets used, meaning that it uses the root-domain.
+!!! warning "Requires manual UI build"
+    This variable requires a manual build of the `n8n-editor-ui` package. You can't use it with the default n8n docker image. The default it`/`, meaning that it uses the root-domain.
```
lorenanda, line 120:
```suggestion
    This variable requires a manual build of the `n8n-editor-ui` package. You can't use it with the default n8n docker image. The default is `/`, meaning that it uses the root-domain.
```

docs/hosting/configuration.md
```diff
@@ -117,11 +116,11 @@ The following environment variables support file input:
...
 
 
-Tells the frontend how to reach the REST API of the backend.
+Tells the frontend how to reach the REST API of the back-end.
```
lorenanda, line 123:
```suggestion
Tells the front end how to reach the REST API of the back end.
```
Edited this based on the MS guide.

StarfallProjects: Heh my hyphen was a compromise - I prefer backend, MS doesn't like it. But you're right, we should stick to the guide we've chosen.

docs/hosting/configuration.md
```diff
@@ -167,20 +166,18 @@ export EXECUTIONS_TIMEOUT_MAX=7200
...
 ```bash
 export N8N_CUSTOM_EXTENSIONS="/home/jim/n8n/custom-nodes;/data/n8n/nodes"
 ```
 
 ### Use built-in and external modules in Function-Nodes
 
-For security reasons, importing modules is restricted by default in the Function-Nodes.
-It is, however, possible to lift that restriction for built-in and external modules by
-setting the following environment variables:
+For security reasons, ithe Function node restricts importing modules. It's possible to lift that restriction for built-in and external modules by setting the following environment variables:
```
lorenanda, line 177:
```suggestion
For security reasons, the Function node restricts importing modules. It's possible to lift that restriction for built-in and external modules by setting the following environment variables:
```

docs/hosting/configuration.md
```diff
@@ -211,43 +205,38 @@ You can find the name of your timezone [here](https://momentjs.com/timezone/).
...
 ### Webhook URL
 
-The webhook URL will normally be created automatically by combining
-`N8N_PROTOCOL`, `N8N_HOST` and `N8N_PORT`. However, if n8n runs behind a
-reverse proxy that would not work. That's because n8n runs internally
-on port 5678 but is exposed to the web via the reverse proxy on port 443. In
-that case, it is important to set the webhook URL manually so that it can be
-displayed correctly in the Editor UI and even more important is that the correct
-webhook URLs get registred with the external services.
+n8n creates the webhook URL by combining `N8N_PROTOCOL`, `N8N_HOST` and `N8N_PORT`. If n8n runs behind a reverse proxy that won't work. That's because n8n runs internally
```
lorenanda, line 217:
```suggestion
n8n creates the webhook URL by combining `N8N_PROTOCOL`, `N8N_HOST` and `N8N_PORT`. If n8n runs behind a reverse proxy, that won't work. That's because n8n runs internally
```
It lures me down the garden path without the comma 😅

lorenanda (approved):

PR #849, opened 2022-05-24
Doc 269 community nodes

StarfallProjects:


docs/integrations/community-nodes/submit-nodes.md
```diff
@@ -0,0 +1,30 @@
...
+
+* View the [starter node](https://github.com/n8n-io/n8n-nodes-starter).
+* Refer to the documentation on [building your own nodes](/integrations/creating-nodes/).
+* Make sure your node follows the [standards](#standards) for community nodes.
+
+## Standards
+
+To make your node available to the n8n community node repository, you must:
+
+* Make sure the package name starts with `n8n-node-` or `@<scope>/n8n-node-`. For example, `n8n-node-weather` or `@weatherPlugins/n8n-node-weather`.
```
StarfallProjects, line 18:
```suggestion
* Make sure the package name starts with `n8n-nodes-` or `@<scope>/n8n-nodes-`. For example, `n8n-nodes-weather` or `@weatherPlugins/n8n-nodes-weather`.
```

docs/integrations/community-nodes/usage.md
```diff
@@ -0,0 +1,12 @@
+# Using community nodes
+
+To use community nodes, you first need to [install](/integrations/community-nodes/installation/){:target=_blank} them.
+
+## Adding community nodes to your workflow
+
+After installing a community node, you can use it like any other node. n8n displays the node in search results in the **Nodes** panel. To show which nodes are community nodes and which are built-in, n8n marks community nodes with a **Package** <span class="inline-image">![Package icon](/_images/common-icons/package.png)</span> icon.
```
lorenanda, line 7:
```suggestion
After installing a community node, you can use it like any other node. n8n displays the node in search results in the **Nodes** panel. n8n marks community nodes with a **Package** <span class="inline-image">![Package icon](/_images/common-icons/package.png)</span> icon.
```

Just a thought: When I read "which are X vs which are Y", I expect the following explanation to have the same structure, like "X are marked with #, Y are marked with *".  If I get only the first part, to me it sounds like something is missing and leaves me thinking "so Y are marked with something else, or are not marked at all?".

docs/integrations/community-nodes/risks.md
```diff
@@ -0,0 +1,17 @@
+# Risks when using community nodes
+
+Using community nodes has some risks. You are installing unverified code from a public source into your n8n instance.
```
lorenanda, line 3:
```suggestion
Using community nodes means you are installing unverified code from a public source into your n8n instance. This has some risks.
```
Just a suggestion: I'd remove the first mention of risks to make it more straightforward, and keep the structure cause (unverified code) - effect (risks).

lorenanda (approved):
Looks good!

PR #872, opened 2022-06-13
Add callout for database command behavior

krynble:
Added this callout to clarify the issue reported here: https://community.n8n.io/t/active-deactive-the-workflow/13332/5


StarfallProjects (approved):
Made some changes

PR #877, opened 2022-06-20
Add label to assign middleware to n8n route

kimtiago:
Add label to assign middleware to n8n route. It's necessary to traefik apply this middleware with header configs to the n8n route.

### DIFF
--- a/docs/hosting/server-setups/docker-compose.md
+++ b/docs/hosting/server-setups/docker-compose.md
@@ -103,6 +103,7 @@ services:
       - traefik.http.middlewares.n8n.headers.SSLHost=${DOMAIN_NAME}
       - traefik.http.middlewares.n8n.headers.STSIncludeSubdomains=true
       - traefik.http.middlewares.n8n.headers.STSPreload=true
+      - traefik.http.routers.n8n.middlewares=n8n@docker
     environment:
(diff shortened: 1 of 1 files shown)


StarfallProjects:
Hi @kimtiago Thanks for the contribution!
Could you give a bit more context? Is this just needed if you want to add middleware - but not essential for a basic setup?

kimtiago:
```
      - traefik.http.middlewares.n8n.headers.SSLRedirect=true
      - traefik.http.middlewares.n8n.headers.STSSeconds=315360000
      - traefik.http.middlewares.n8n.headers.browserXSSFilter=true
      - traefik.http.middlewares.n8n.headers.contentTypeNosniff=true
      - traefik.http.middlewares.n8n.headers.forceSTSHeader=true
      - traefik.http.middlewares.n8n.headers.SSLHost=${DOMAIN_NAME}
      - traefik.http.middlewares.n8n.headers.STSIncludeSubdomains=true
      - traefik.http.middlewares.n8n.headers.STSPreload=true
```

these labels are creating the middleware n8n and setting all these options in this middleware, but only creating isn't enough to route n8n use this...
without the label I committed, these options will not be applied...

so, don't make sense all this options here if n8n don't use this

StarfallProjects:
Thank you. Merged.

PR #892, opened 2022-07-14
Draft GCP hosting guide

ChrisChinchilla:


docs/hosting/server-setups/google-cloud.md
```diff
@@ -0,0 +1,146 @@
+# Hosting n8n on Google Cloud
+
+This hosting guide shows you how to self-host n8n on Google Cloud (GCP). It uses n8n with Postgres as a database backend using Kubernetes to manage the necessary resources and reverse proxy.
+
+## Prerequisites
+
+- [The gcloud command line tool](https://cloud.google.com/sdk/gcloud/){:target="_blank" .external-link}
```
StarfallProjects, line 7: Does it make sense to set this as a prerequisite if we don't use it in the steps?

docs/hosting/server-setups/google-cloud.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Enable the Kubernetes Engine API
+
+GKE isn't enabled by default, search for "Kubernetes" in the top search bar and select "Kubernetes Engine" from the results.
+
+Enable the Kubernetes Engine API by clicking the **Enable** button.
+
+## Create a cluster
+
+From the GKE service page, click the **Clusters** menu item and then the **CREATE** button. Make sure you select the "Standard" cluster option, n8n doesn't work with an "Autopilot" cluster.
```
StarfallProjects, line 29: There's then a config page. I assume you skipped it as the defaults are ok, but it would be good to reassure the reader about this.

ChrisChinchilla: They had also changed the UI twice in the process of writing this, but I'll add something a little vague around "accept defaults" to cover bases.

```suggestion
From the GKE service page, click the **Clusters** menu item and then the **CREATE** button. Make sure you select the "Standard" cluster option, n8n doesn't work with an "Autopilot" cluster. You can leave the cluster configuration on defaults unless there's anything specifically you need to change, such as location.
```

docs/hosting/server-setups/google-cloud.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Set Kubectl context
+
+The remainder of the steps in this guide require you to set the GCP instance as the Kubectl context. You can find the connection details for a cluster instance by opening its details page and then the **CONNECT** button. The resulting code snippet shows a connection string for the gcloud CLI tool. Paste and run that code snippet into a terminal to change your local Kubernetes settings to use the new gcloud cluster.
+
+## Setup DNS
+
+n8n typically operates on a subdomain. Create a DNS record with your provider for the subdomain and point it to a static IP address of the instance.
+
+If the instance doesn't already have a static IP address, you can assign one to it by editing the instance, and changing the network interface from "Ephemeral" to "Static".
```
StarfallProjects, line 39: I got fairly lost at this point:

- I think I found the right place to configure IP addresses for the cluster(s), but it would be good to confirm this
- This tutorial suggests changing the network interface to static, but it's not clear how to do that. Googling found [this guide](https://www.howtogeek.com/devops/how-to-assign-a-static-ip-address-to-your-google-cloud-platform-instances/) which suggests we need to go through a process to reserve a static external IP?
- While investigating this, I also discovered that I seem to have created a cluster with 3 VM instances (I assume intentional) Would we need to set up static IP and DNS for all of them?

docs/hosting/server-setups/google-cloud.md
```diff
@@ -0,0 +1,146 @@
...
+
+<!-- TODO: Kubernetes handles? -->
```
StarfallProjects, line 41: Should this be in?

ChrisChinchilla: Nope

```suggestion
```

docs/hosting/server-setups/google-cloud.md
ChrisChinchilla: Hmm, this was another area where they moved the interface around a lot, let me take another look. But I'm not entirely sure what you mean by 3 VM instances, maybe it's three nodes, which is expected, but shouldn't really be any VMs… Got a screenshot?

StarfallProjects: image

I ended up in there because the guide I found (linked above) suggested configuring IP addresses here rather than someplace in the kubernetes engine settings. So I may have been entirely in the wrong place.

ChrisChinchilla: The impression I actually get is that it now leverages Kubernetes config to expose an IP address, not anything in configuration, but still need a way to easily describe how people can find that. Will add.

ChrisChinchilla: And I see what's happened… You're using "Google Compute Engine" not "Google Kubernetes Engine". Love that naming!

StarfallProjects: That was what was suggested in the guide - I couldn't find the equivalent in GKE. But it did seem weird! But those are the nodes belonging to the GKE cluster (I think)

ChrisChinchilla: @StarfallProjects As far as I could tell GKE now reserves address if you use the appropriate Kubernetes resource. But it's super vague, so I opted to mention where to find what i think is the "reserved" address and added where people can find more details, which means n8n doesn't have to keep up with what Google might change too.

StarfallProjects: @ChrisChinchilla  I'm seeing Services & Ingress on the left menu (when in the Kubernetes Engine page in my project). I'm unsure where to find it following the instructions (Endpoints column?)

When I click it in the left menu, I get more setup steps. Looking at the guide you linked, I think people need to go through the "exposing your application" steps, before configuring DNS.

Can you confirm / update as needed?

PR #916, opened 2022-08-09
fix google ads issues

StarfallProjects:


that-one-tom (approved):
Looking good, thanks so much!

PR #935, opened 2022-08-18
Added warning for scopes to be added to the app in Hubspot

bramkn:
If the scopes aren't exactly like those listed in the docs this can cause errors during authentication of the OAuth2 credentials.
I simply granted all, and got the error: 'Insufficient scopes were provided' 
Which makes no sense of course.


docs/integrations/builtin/credentials/hubspot.md
```diff
@@ -27,6 +27,10 @@ Create a [HubSpot](https://www.hubspot.com/) account.
 6. Use the provided **Client ID** and the **Client secret** with your HubSpot OAuth2 API credentials in n8n.
 7. If you are using the [HubSpot Trigger](/integrations/builtin/trigger-nodes/n8n-nodes-base.hubSpotTrigger/) node, copy the **App ID** along with the information in the previous step.
 8. Copy your OAuth Callback URL from the 'Create New Credentials' screen in n8n and paste in the **Redirect URL** section.
+
+!!! warning "Exact scope needed"
```
StarfallProjects, line 31:
```suggestion
!!! note "Exact scope needed"
```

StarfallProjects: We try to save 'warning' for really major things: destructive actions, security topics etc. This addition is important, but not dangerous, so I think 'note' is better here.

bramkn: Ok, sounds good with that explanation.

StarfallProjects:
Thank you for the contribution! 🎉

PR #947, opened 2022-08-31
[bitly] fix bitly node text in example usage

bhanuprasadcherukuvada:
ref - 
#948

A small fix. In the example usage of the `Bitly` node, It was mentioned to enter the credentials of the `Twilio` node, but instead, it should be the `Bitly` node.


StarfallProjects:
Thank you! 🎉

PR #966, opened 2022-09-27
google creds and expanding text

StarfallProjects:


StarfallProjects:
Related dev PR: https://github.com/n8n-io/n8n/pull/4381

PR #968, opened 2022-10-01
Update quickstart for new expressions interface

rosano:
There is no longer 'Parameter options', just a slider that says 'Fixed | Expression'.


StarfallProjects:
Great, thank you! 🙏

PR #969, opened 2022-10-01
 Update longer introduction for new expressions interface

rosano:
There is no longer 'Parameter options', just a slider that says 'Fixed | Expression'.


StarfallProjects:
And thanks again! 🙏

PR #970, opened 2022-10-02
Use friendly key names

rosano:
I find these easier to read and more instructive for people less familiar with JSON.

### DIFF
--- a/docs/courses/level-one/chapter-5/chapter-5.5.md
+++ b/docs/courses/level-one/chapter-5/chapter-5.5.md
@@ -29,15 +29,15 @@ In n8n, the data that is passed between nodes is an array of objects with the fo
    	 // Any kind of JSON data is allowed. So arrays and the data being deeply nested is fine.
    	 json: { // The actual data n8n operates on (required)
    		 // This data is only an example it could be any kind of JSON data
-   		 jsonKeyName: 'keyValue',
-   		 anotherJsonKey: {
-   			 lowerLevelJsonKey: 1
+   		 alfa: 'bravo',
+   		 charlie: {
+   			 delta: 1
    		 }
    	 },
    	 // Binary data of item. The most items in n8n do not contain any (optional)
    	 binary: {
-   		 // The key-name "binaryKeyName" is only an example. Any kind of key-name is possible.
-   		 binaryKeyName: {
+   		 // The key-name "alfa" is only an example. Any kind of key-name is possible.
+   		 alfa: {
    			 data: '....', // Base64 encoded binary data (required)
    			 mimeType: 'image/png', // Optional but should be set if possible (optional)
    			 fileExtension: 'png', // Optional but should be set if possible (optional)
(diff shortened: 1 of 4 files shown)


StarfallProjects:
Friendlier key names is a good suggestion. I'm concerned this might confuse people - I found myself trying to work out where they're from, before remembering the NATO phonetic alphabet is a thing. We probably want to either keep the original somewhat descriptive key names, or choose something even friendlier as an example (use of names in docs examples is a bigger project . . .)

What are your thoughts?

rosano:
Could also be 'apple', 'banana', 'carrot', 'dog'… I think some beginners might feel uncomfortable changing 'jsonKeyName'  because they thing you need to say 'KeyName' or something. And 'foo' / 'bar' is more legible to programmers. When it's clear that the names are arbitrary and not special keywords, I think people more easily tweak and play.

StarfallProjects:
Agreed. Would you be willing to update the example to use fruit and vegetables? I think that's a reasonable solution for now.

rosano:
Done.

I'm a bit confused about how to resolve the conflict as it seems that the whole `Data structure` section was actually not there at the beginning? 🧐 https://github.com/n8n-io/n8n-docs/blob/main/docs/integrations/builtin/core-nodes/n8n-nodes-base.function.md Am I understanding correctly?

StarfallProjects:
Ah I think you started on this before we merged a fairly large chunk of work. Data structure info lives here now: https://github.com/n8n-io/n8n-docs/blob/main/docs/data/data-structure.md https://docs.n8n.io/data/data-structure/ It looks like we link to it from the function node page.

docs/data/data-structure.md
```diff
@@ -9,16 +9,16 @@ In n8n, all data passed between nodes is an array of objects. It has the followi
...
+			"carrot": {
+				"dill": 1
 			}
 		},
 		// For binary data:
 		// Wrap each item in another object, with the key 'binary'
 		"binary": {
 			// Example data
-			"binaryKeyName": {
+			"apple": {
```
StarfallProjects, line 21:
```suggestion
			"apple-picture": {
```

Think this makes more sense / is a bit more informative

docs/courses/level-one/chapter-5/chapter-5.5.md
```diff
@@ -29,15 +29,15 @@ In n8n, the data that is passed between nodes is an array of objects with the fo
...
+   		 carrot: {
+   			 dill: 1
    		 }
    	 },
    	 // Binary data of item. The most items in n8n do not contain any (optional)
    	 binary: {
-   		 // The key-name "binaryKeyName" is only an example. Any kind of key-name is possible.
-   		 binaryKeyName: {
+   		 // The key-name "apple" is only an example. Any kind of key-name is possible.
+   		 apple: {
```
StarfallProjects, line 40:
```suggestion
   		 apple-picture: {
```

StarfallProjects (changes requested):
Small suggestion for the binary data key, to make it a bit clearer what's going on there

rosano:
Looks real friendly 👍🏽

rosano (commented):
Looks real friendly 👍🏽

PR #989, opened 2022-10-21
Removed useless numbers (typos)

albertopasqualetto:
Those numbers are also in the images

### DIFF
--- a/docs/integrations/builtin/app-nodes/n8n-nodes-base.microsoftToDo.md
+++ b/docs/integrations/builtin/app-nodes/n8n-nodes-base.microsoftToDo.md
@@ -57,7 +57,7 @@ In the screenshot below, you will notice that the node creates a new task in Mic
 
 ![Create a task with the Microsoft To Do node](/_images/integrations/builtin/app-nodes/microsofttodo/microsofttodo_node.png)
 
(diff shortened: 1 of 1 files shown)


StarfallProjects:
@albertopasqualetto thanks for the contribution! However, these are not 'useless' numbers: this example contains three different todo nodes, so the numbers in their names differentiate them.

Ideally this example would be overhauled to use more meaningful names, as well as follow the [style guide](https://github.com/n8n-io/n8n-docs/wiki/Styles). If you feel like taking that on, that'd be awesome!

Closing this for now.

albertopasqualetto:
Maybe I am not understanding something but in my version of n8n (updated at time of this PR) the node is only one and it can be set to get, update,...
Since get and update already differentiated them, the numbers were unnecessary.
What am I not getting?

StarfallProjects:
In the example you updated, three copies of the node are added to the workflow, hence them needing different names. You can't have multiple nodes with the same name on the canvas. If you follow the tutorial, you will see these are the names automatically created when you add more nodes.

However, the example is out of date in other ways now, so probably does need a rewrite or removal.

albertopasqualetto:
I understand now

PR #1017, opened 2022-11-21
release notes and google sheets update

StarfallProjects:


michael-radency (approved):
in read rows section we could mention filters and option to output only first match(default) or all matches

StarfallProjects:
> in read rows section we could mention filters and option to output only first match(default) or all matches

Added.

PR #1054, opened 2022-12-24
add mutate example for graphql

leangseu:


StarfallProjects:
Hi @leangseu thank you for this. However, as far as I can see, this is just a generic example of how to do a mutation in GraphQL?

I think it might be a better approach to:

1. Update the doc according the the [template](https://github.com/n8n-io/n8n-docs/blob/main/document-templates/core-nodes.md)
2. Remove the current examples
3. Add a link to a good GraphQL learning resource / to good query examples

Are you up for doing this?

PR #1065, opened 2023-01-10
Doc 460 jmespath

StarfallProjects:


docs/code-examples/expressions/jmespath.md
```diff
@@ -225,3 +174,53 @@ Use multiselect list to get the first and last names and create new lists contai
...
 
+
+### An alternative to arrow functions
+
+You can't use arrow functions in expressions. This means you can't easily do things like looking for an item with a specific key.
+
+For example, given the below input data, you could do a search like "find the item with the name Lenovo and tell me their category ID."
+
+```js
+return[
```
janober, line 185: I would return both items underneath the `json` property. It for sure works like that, as n8n dos now add it automatically, but it would then make the example that follows clearer (as it also uses `json`).

PR #1076, opened 2023-01-16
upgrade theme and enable breadcrumbs

StarfallProjects:


StarfallProjects:
Also, have filed a feature suggestions to hide breadcrumbs on some page: https://github.com/squidfunk/mkdocs-material/issues/4875

PR #1086, opened 2023-01-21
Fixed Docker Compose Commands

Nabeelshar:
There is mistake in docker commands instead of docker-compose it was docker compose which is not not correct command.
Locations:
hosting/installation/docker/
hosting/installation/server-setups/docker-compose/

### DIFF
--- a/docs/hosting/installation/docker.md
+++ b/docs/hosting/installation/docker.md
@@ -139,13 +139,13 @@ If you've running n8n using a Docker Compose file, follow the below mentioned st
 
 ```sh
 // Pull latest version
-docker compose pull
+docker-compose pull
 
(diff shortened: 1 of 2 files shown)


StarfallProjects:
Hi @Nabeelshar, thanks for the PR. However, this was a deliberate change recently (from `docker-compose` to `docker compose`)

`docker-compose` was the command in Docker Compose V1. Docker Compose V2 launched in April 2022 (see [announcement](https://www.docker.com/blog/announcing-compose-v2-general-availability/) for more info) V1 is now deprecated, and from April 2023 Docker will remove the ability to opt out of V2. So our docs reflect the [V2 commands](https://docs.docker.com/compose/reference/).

PR #1100, opened 2023-02-01
Use valid JSON + working query parameter example

that-one-tom:
This PR replaces the current example from https://docs.n8n.io/integrations/builtin/app-nodes/n8n-nodes-base.postgres/#use-query-parameters using expressions with an example using a static field name to match the current functionality of the Postgres node.


that-one-tom:
Sounds good, but I am afraid I don't have the time for it. Giulio discovered this error so I figured I'll just raise a PR to fix it. If you'd rather re-do this page feel free to simply close it :)

StarfallProjects:
Nah let's merge the fix and revisit this node when there's time. Thanks for the PR :-)

StarfallProjects (approved):

PR #1101, opened 2023-02-01
Add page on isolating n8n

that-one-tom:
Adds a page explaining how to isolate self-hosted n8n instances. This is one of our action items following the recent Azure networking problems affecting the n8n server connectivity.


StarfallProjects:
Made some git mess and this is now here: https://github.com/n8n-io/n8n-docs/pull/1102

PR #1132, opened 2023-02-22
app-node-intros

tiseysoft:


ppolinaua:
@tiseysoft one (hopefully) last comment :)

Bannerbear node
Should be:
The Bannerbear node (to add The)

PR #1133, opened 2023-02-22
Update update links

OpenSourceSimon:


StarfallProjects:
Hi @OpenSourceSimon thanks for the contribution. However, we don't link direct to headings as a general rule.

PR #1142, opened 2023-02-24
Hunter node

tiseysoft:


ppolinaua:
@tiseysoft with this node I'd say "email addresses" not just emails, otherwise it seems like this node gets and verifies emails, but it does it with email addresses

ppolinaua:
@StarfallProjects ✅

PR #1147, opened 2023-02-25
Jira

tiseysoft:


ppolinaua:
@tiseysoft a wrong link - should be this one https://n8n.io/integrations/jira-software/

PR #1149, opened 2023-02-25
Kitemaker

tiseysoft:


ppolinaua:
@tiseysoft list of operations also doesn't make sense: it's better to say ".... a wide range of Kitemaker features, including retrieving data on the organizations, spaces and users as well as creating, getting and updating work items'

I'd encourage you to double check if the listed operations make sense because a reader should understand what's possible to do with these operations: what item does what

PR #1173, opened 2023-02-28
Matrix

tiseysoft:


ppolinaua:
@tiseysoft I am afraid it's again a random set of operations that doesn't make sense: 'creating, updating, inviting, and getting accounts, media, and rooms'
It reads like you can invite rooms and media and etc. Please, let's use correct item with its correct operation. Better to list less items but correct ones

ppolinaua:
@StarfallProjects ✅

PR #1199, opened 2023-03-01
Openai

tiseysoft:


ppolinaua:
@tiseysoft this part of the sentence is good but the order of words is weird: 'including creating images, and completing, as well as moderating, and editing texts.'

PR #1278, opened 2023-03-15
Incorrect node expression

Chaoscontrol:
The Unique ID is not being mapped with this expression. After checking, headerParametersUi doesn't exist in my version and it's named headerParameters instead. It worked when I fixed it.


StarfallProjects:
Thank you for the contribution 🙏

PR #1308, opened 2023-03-19
uproc

tiseysoft:


ppolinaua:
@tiseysoft Seems like this node can offer way toooo more :) So I'd suggest listing operations names such as 'including getting audio, communication data, company, finance and product information'

ppolinaua:
@StarfallProjects ✅

PR #1349, opened 2023-04-13
add release notes for 224, new totp node, postgres update

StarfallProjects:


docs/integrations/builtin/app-nodes/n8n-nodes-base.postgres.md
```diff
@@ -99,10 +54,13 @@ For example, you want to find a person by their email address. Given the followi
 You can write a query like:
 
 ```sql
-SELECT * FROM users WHERE email = $1;
+SELECT * FROM $1:table_name WHERE email = $2;
```
StarfallProjects, line 57:
```suggestion
SELECT * FROM $1:name WHERE email = $2;
```

PR #1350, opened 2023-04-14
Update log levels

dacook:
I received an error when starting the service, which provided the following names:

    ["error","warn","info","verbose","debug","silent"]: value was "warning"

Behaviour of `silent` has been inferred from the name..

**Please review to double-check that this is correct.**


StarfallProjects:
Thank you! 🙏

PR #1364, opened 2023-04-20
Add proprietary license text

janober:


StarfallProjects:
@janober I think you may be on quite an out of date version of the docs? That reference section hasn't existed for a while (license info is now here: https://docs.n8n.io/choose-n8n/faircode-license/)

janober:
Thanks! Yes, probably had to to with me using "master" and not "main".

Created now a new PR. #1365

PR #1366, opened 2023-04-22
update apt-get update for hetzner

NoahBohme:
The default for Hetzner Cloud with Docker-CE is now Ubuntu 22.04 - Therefore the update command should be apt-get update

Also added apt-get upgrade.

### DIFF
--- a/docs/hosting/installation/server-setups/hetzner.md
+++ b/docs/hosting/installation/server-setups/hetzner.md
@@ -32,7 +32,8 @@ The rest of this guide requires you to log in to the server using a terminal wit
 The Hetzner Docker app image doesn't have Docker compose installed. Install it with the following commands:
 
 ```shell
-apt get update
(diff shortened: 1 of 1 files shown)


Joffcom:
@NoahBohme would you also accept just `apt update` and `apt upgrade` as these are the preferred commands for Ubuntu?

NoahBohme:
Hi @Joffcom 
I was not able to run apt get update when booting up a Docker CE on Hetzner Cloud

Joffcom:
Hey @NoahBohme,

That part doesn't surprise me and I think that was a typo that was not picked up on before, I believe the correct command should be `apt update` and `apt upgrade` without the other part. From what I remember while `apt-get` will work just `apt` has been preferred for a while.

Joffcom:
@NoahBohme you could have updated the same PR if it was easier.

PR #1370, opened 2023-04-24
Add capabilities to Notion credential

Joffcom:
For reference...
image

### DIFF
--- a/docs/integrations/builtin/credentials/notion.md
+++ b/docs/integrations/builtin/credentials/notion.md
@@ -21,5 +21,12 @@ Notion provides access using API keys. To get and use the API key, you need:
 
 Follow Step 1 in the [Notion integration guide](https://developers.notion.com/docs/create-a-notion-integration){:target=_blank .external-link} to create your integration and get your Notion integration token (API key).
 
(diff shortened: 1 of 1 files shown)


docs/integrations/builtin/credentials/notion.md
```diff
@@ -21,5 +21,12 @@ Notion provides access using API keys. To get and use the API key, you need:
 
 Follow Step 1 in the [Notion integration guide](https://developers.notion.com/docs/create-a-notion-integration){:target=_blank .external-link} to create your integration and get your Notion integration token (API key).
 
+For integration capabilities we recommend enabling the following for the node to fully function:
```
StarfallProjects, line 24:
```suggestion
On the **Capabilities** screen, enable the following:
```

StarfallProjects (changes requested):
Suggest rephrasing to be more direct/firm. And let's just assume they need the user resource, because people will trip themselves up otherwise.

docs/integrations/builtin/credentials/notion.md
Joffcom: 👍🏻

PR #1444, opened 2023-05-31
0.231.0 release note

StarfallProjects:


docs/integrations/creating-nodes/build/reference/ui-elements.md
```diff
@@ -564,4 +564,95 @@ The HTML editor allows users to create HTML templates in their workflows. The ed
...
+	Matching is the process of using column names to identify the row(s) to update.
+	Mapping is the process of setting the input data to use as values when updating row(s).
+
+
+```js
+{
+	displayName: 'Columns',
+	name: 'columns', // The name used to reference the element UI within the code
+	type: 'resourceMapper', // The UI element type
+	default: {
```
MiloradFilipovic, line 581: Can you also add the `value: null` should be used in `default`(as mentioned [here](https://www.notion.so/n8n/Resource-Mapper-Component-792800eddab54aa59d061039490eb50b?pvs=4#7bc84e3835974bc8bf95630c7c10068c))?

docs/integrations/creating-nodes/build/reference/ui-elements.md
```diff
@@ -564,4 +564,95 @@ The HTML editor allows users to create HTML templates in their workflows. The ed
...
+		// mappingMode can be defined in the component (mappingMode: 'defineBelow')
+		// or you can attempt automatic mapping (mappingMode: 'autoMapInputData')
+		mappingMode: 'defineBelow',
+	},
+	required: true,
+	typeOptions: {
+		resourceMapper: {
```
MiloradFilipovic, line 588: I think it would be useful to know which one of these options are required and which ones are not. I would either add it to comments or show the [interface](https://www.notion.so/n8n/Resource-Mapper-Component-792800eddab54aa59d061039490eb50b?pvs=4#5479f17a5d5f4b4bbfcea926355319c6) they implement.

docs/integrations/creating-nodes/build/reference/ui-elements.md
```diff
@@ -564,4 +564,95 @@ The HTML editor allows users to create HTML templates in their workflows. The ed
 
 Refer to [`Html.node.ts`](https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/Html/Html.node.ts){:target=_blank .external-link} for a live example.
 
+## Resource mapper
 
+If your node performs insert, update, or upsert operations, you need to send data from the node in a format supported by the service you're integrating with. A common pattern is to use a Set node before the node that sends data, to convert the data to match the schema of the service you're connecting to. The resource mapper UI component provides a way to get data into the required format directly within the node, rather than using a Set node.
```
MiloradFilipovic, line 569: I think we should mention that _getting the data in the right format_ means that RMC will validate input against the schema and also cast it into expected type.

PR #1479, opened 2023-06-26
Add new node docs

Joffcom:
Waiting to find out if "Crowd Dev" is correct or if they would always prefer "crowd.dev" which is how they refer to themselves.


Joffcom:
... they do prefer crowd.dev, Will update in the morning

PR #1490, opened 2023-07-05
Mongo db credentials update

gandreini:
I added some updates on how to configure Mongo DB Atlas.
Currently added to a section: **Connecting to Mongo DB Atlas**
I don't know Mongo DB very well and I'm not sure if these guidelines apply also to other versions of Mongo DB.


gandreini:
I replaced "using OAuth" with "Using Connection String or Values" and added some generic information about how to connect to any type of MongoDB. Added link to connection string details.

StarfallProjects (changes requested):
@gandreini couple of things you could quickly fix:

- UI text should be bold but not italix
- Some of the screenshots are unclear/details too small to read.
- All info must be in text as well as screenshot (accessibility)

PR #1567, opened 2023-08-25
Bitbucket Creds copy tweaks

gandreini:
Minor tweaks to the page to better guide the user to retrieve the info from the Bitbucket personal configuration.


StarfallProjects:
Thanks! 🙏

PR #1559, opened 2023-08-16
Update deployment guides to use volumes

Joffcom:
Docker Caddy Deployment PR: https://github.com/n8n-io/n8n-docker-caddy/pull/11

PR #1570, opened 2023-08-28
Update docker.md

that-one-tom:
The existing `docker run` command provided to start n8n with tunnel won't work anymore with n8n version 1 (or on Windows) as the `n8n` part of the command has to be committed.

This PR ensures the example command works with n8n version 1 and when using Windows while maintaining compatibility with Mac and Linux.


Joffcom:
This is also covered in #1559

StarfallProjects (approved):

PR #1577, opened 2023-08-30
fix command to start n8n in queue mode

rgomezcasas:
The previous command was failing with an `Error: command n8n not found` error


StarfallProjects:
@Joffcom is this part of what's covered by your PR as well? Want to merge this into that, or?

Joffcom:
This is indeed covered in my changes although the syntax here is a bit different and includes the full path which isn't needed.

StarfallProjects:
In that case I'm going to close this - but thank you for the contribution @rgomezcasas !

rgomezcasas:
Okkk, thanks!

PR #1603, opened 2023-09-18
update DB doc

StarfallProjects:


docs/hosting/architecture/database-structure.md
```diff
@@ -19,20 +19,36 @@ To inspect the n8n database, you can use [DBeaver](https://dbeaver.io){:target=_
...
+Stores the history of a SAML connection. [TODO - confirm]
+
 ### credentials_entity
 
 Stores the credentials used to authenticate with integrations.
 
-### eventdestinations_entity
+### event_destinations
 
 Enterprise only.
```
Joffcom, line 36: This table does still exist, This part may not be needed.

docs/hosting/architecture/database-structure.md
```diff
@@ -19,20 +19,36 @@ To inspect the n8n database, you can use [DBeaver](https://dbeaver.io){:target=_
...
 
 Contains the destination configurations for [Log streaming](/log-streaming/).
 
+### execution_data
+
+Links executions to workflow data. [TODO - confirm]
```
Joffcom, line 42: This is the actual execution data and contains the workflow at the time of running and the execution data

docs/hosting/architecture/database-structure.md
StarfallProjects: I can't see it with DBeaver? It's not in the diagram it generated either?

Joffcom: @StarfallProjects that is interesting, I can see it in my Postgres database. I would be surprised if we made a table when the license is populated 🤔

StarfallProjects: Let us never speak of this again

Joffcom (approved):

PR #1633, opened 2023-10-04
update data pruning documentation

tomi:
Only executions that are in an end-state are pruned and the max count is not always exact.

Related PR: https://github.com/n8n-io/n8n/pull/7333

### DIFF
--- a/docs/hosting/scaling/execution-data.md
+++ b/docs/hosting/scaling/execution-data.md
@@ -4,9 +4,7 @@ contentType: howto
 
 # Execution data
 
-
 Depending on your executions settings and volume, your n8n database can quickly grow in size and eventually run out of storage.
-
 
 To avoid this, n8n recommends that you don't save unnecessary data, and enable pruning of old executions data.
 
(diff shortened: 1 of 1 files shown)


docs/hosting/scaling/execution-data.md
```diff
@@ -58,13 +55,11 @@ n8n:
...
 
-You can enable data pruning to automatically delete executions after a given time period. If you don't set `EXECUTIONS_DATA_MAX_AGE`, 336 hours (14 days) is the default.
-
-You can choose to prune executions data before the time set in `EXECUTIONS_DATA_MAX_AGE`, using `EXECUTIONS_DATA_PRUNE_MAX_COUNT`. This sets a maximum number of executions to store in the database. Once you reach the limit, n8n starts to delete the oldest execution records. This can help with database performance issues, especially if you use SQLite.
+You can enable data pruning to automatically delete finished executions after a given time period. If you don't set `EXECUTIONS_DATA_MAX_AGE`, 336 hours (14 days) is the default.
 
+You can choose to prune finished executions data before the time set in `EXECUTIONS_DATA_MAX_AGE`, using `EXECUTIONS_DATA_PRUNE_MAX_COUNT`. This sets a maximum number of executions to store in the database. Once you reach the limit, n8n starts to delete the oldest execution records. This can help with database performance issues, especially if you use SQLite. This number is however not exact, and in some cases there might be more executions left in the database.
```
StarfallProjects, line 62: "This number is however not exact, and in some cases there might be more executions left in the database." - this is vague. Do you mean n8n doesn't always stick to the maximum set? (I'm guessing if pruning only runs every so often, it could lead to the DB growing beyond the max temporarily? At any rate, can we give a reason?)

PR #1645, opened 2023-10-10
Doc 633 workflow history

StarfallProjects:


docs/workflows/history.md
```diff
@@ -0,0 +1,46 @@
...
+
+!!! info "Feature availability"
+	* Full workflow history is available on Enterprise Cloud and Enterprise Self-hosted.
+	* Versions from the last five days are available for Cloud Pro users.
+
+Use workflow history to view and restore previous versions of your workflows. 
+
+## Understand workflow history
+
+n8n creates a new version when you:
```
StarfallProjects, line 17:
```suggestion
n8n creates a new version when you: [TEST]
```

jcn8n (approved):
Feature now available as part of 1.15.1 so should be good to go making these docs live. All make sense to me, might be some small tweaks to follpw-up re specific retention for cloud users (Both pro & Enterprise) but I can commit those later if needed.

PR #1689, opened 2023-10-23
Add note about pruning in S3 mode

ivov:
Ref https://github.com/n8n-io/n8n/pull/7482

### DIFF
--- a/docs/hosting/external-storage.md
+++ b/docs/hosting/external-storage.md
@@ -38,7 +38,7 @@ Create and configure a bucket following the [AWS documentation](https://docs.aws
 }
 ```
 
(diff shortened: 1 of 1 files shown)


docs/hosting/external-storage.md
```diff
@@ -38,6 +38,8 @@ Create and configure a bucket following the [AWS documentation](https://docs.aws
 }
 ```
 
+Set a bucket-level lifecycle configuration so that S3 automatically deletes old binary data. n8n delegates pruning of binary data to S3, so setting a lifecycle configuration is required unless you would like to preserve binary data indefinitely.
```
StarfallProjects, line 41:
```suggestion
Set a bucket-level lifecycle configuration so that S3 automatically deletes old binary data. n8n delegates pruning of binary data to S3, so setting a lifecycle configuration is required unless you want to preserve binary data indefinitely.
```

StarfallProjects:
@ivov  great thanks! Feel free to merge this whenever the dev work is merged.

PR #1741, opened 2023-11-18
🐛 Fix invalid TypeScript for resourceLocator

pemontto:
The current `resourceLocator` sample code is incorrect. This PR fixes the example and formats with Prettier based on the n8n repo settings.


Joffcom (approved):
Looks good thanks @pemontto

PR #1757, opened 2023-11-29
Add notice about trigger privilege

Joffcom:


docs/integrations/builtin/trigger-nodes/n8n-nodes-base.postgrestrigger.md
```diff
@@ -31,7 +31,6 @@ n8n provides an app node for Postgres. You can find the node docs [here](/integr
 
 View [example workflows and related content](https://n8n.io/integrations/postgres-trigger/){:target=_blank .external-link} on n8n's website.
 
-
-
-
-
+/// note | Permissions
+You will need to make sure your database user has the `TRIGGER` privilege, You can find more information on this in the Postgres documentation [here](https://www.postgresql.org/docs/current/ddl-priv.html#DDL-PRIV-TRIGGER){:target=_blank .external-link}.
```
StarfallProjects, line 35:
```suggestion
Make sure your database user has the `TRIGGER` privilege, You can find more information on this in the Postgres documentation [here](https://www.postgresql.org/docs/current/ddl-priv.html#DDL-PRIV-TRIGGER){:target=_blank .external-link}.
```

StarfallProjects:
@Joffcom is it worth adding a link or more info in the credentials doc so that users can find out what the "correct privileges" are?

StarfallProjects (approved):

Joffcom:
@StarfallProjects I did think about maybe listing the ones we would need without using the query option which cover most options. I may also do the same thing for the MySQL node and the MSSQL node just to be complete.

StarfallProjects:
@Joffcom are you still planning to do this? 😅

Joffcom:
@StarfallProjects I forgot about this 😓

PR #1805, opened 2023-12-23
Mark Compatibility with Service Account in BigQuery as supported

antonialoytorrens:
This PR updates the documentation to mark the Service Account authentication method as supported for the Google BigQuery node.

I have personally been using the Service Account method for authentication with the BigQuery node and have found it to be fully functional and reliable, without encountering any issues.

For further reference, please see [the relevant section of the codebase](https://github.com/n8n-io/n8n/blob/f50fc8443e3359b0a3b0ca991b4be878aa52e606/packages/nodes-base/nodes/Google/BigQuery/v2/transport/index.ts#L21-L25) where the Service Account authentication method is implemented as a viable option.


StarfallProjects:
Thank you! 🙏

PR #1812, opened 2024-01-02
feat: Add tags operations and filter workflows by name in Public API

jburgers-chakray:
Hello,

This change is related to the PR: [https://github.com/n8n-io/n8n/pull/8202](https://github.com/n8n-io/n8n/pull/8202)

Thanks!


StarfallProjects:
Hi @jburgers-chakray thanks for this.
What we usually do for the API reference is grab the built spec file when it changes (rather than maintaining a separate spec file in the docs) I realise this is not clear from looking at the repo though.

I'll leave this PR open. Once the dev work is merged, one of us can get the spec file and switch it out.

jburgers-chakray:
Hi @StarfallProjects,

That makes sense. Thanks!

StarfallProjects:
Thanks for this! When the alternative dev PR went live I grabbed the new OpenAPI file from there. Closing this, but please do let me know if there are any issues with this: https://docs.n8n.io/api/api-reference/

PR #1825, opened 2024-01-10
Doc 727 vars read onl

StarfallProjects:


docs/code/variables.md
```diff
@@ -44,3 +45,5 @@ $vars.<variable-name>
 All variables are strings.
 
 During workflow execution, n8n replaces the variables with the variable value. If the variable has no value, n8n treats its value as `undefined`. Workflows don't automatically fail in this case.
+
+Variables are read-only. You must use the UI to change the values. If you need to set and access custom data within your workflow, use [Custom executions data](/workflows/executions/custom-executions-data/).
```
krynble, line 49: Maybe we could suggest `https://docs.n8n.io/code/cookbook/builtin/get-workflow-static-data/` instead - both work but the "purpose" of execution data is to be searchable but it's not persisted across executions.

StarfallProjects: Good point, have changed it. Ok for this to go now?

krynble (approved):
Thanks a lot @StarfallProjects

PR #1847, opened 2024-01-18
update edit fields node and add new component

StarfallProjects:


docs/integrations/creating-nodes/build/reference/ui-elements.md
```diff
@@ -412,6 +412,24 @@ options: [
 
 ![Filter](/_images/integrations/creating-nodes/filter.png)
 
+
+## Assignment collection (drag and drop)
+
+Use the drag and drop component when you want users to pre-fill multiple parameters with a single drag interaction.
```
elsmr, line 418: The assignment component is a bit more specific. It's always a Name and a Value field (and optionally a type)

elsmr (approved):

docs/integrations/creating-nodes/build/reference/ui-elements.md
StarfallProjects: Updated, thank you!

PR #1885, opened 2024-02-07
Additions to multi-main setup docs

ivov:


docs/hosting/scaling/queue-mode.md
```diff
@@ -181,36 +181,40 @@ n8n worker --concurrency=5
...
+In a multi-main setup, there are two kinds of `main` processes:
 
-The leader `main` process reports to Redis, setting a short-lived key. If the leader ever fails to renew the key, for example because it crashed or because its event loop became overly busy, n8n designates one of the follower `main` processes as the new leader to take over the former leader's responsibilities. If the previous leader becomes responsive again, it becomes a follower.
+- **followers**, which run **regular tasks**, and
+- the **leader**, which runs **both regular and at-most-once tasks**.
+
+### Leader designation
+
+In a multi-main setup, all `main` processes attempt to set a short-lived key on Redis - the `main` process that first sets the key becomes the leader and will keep renewing the key to remain the leader. If the leader ever fails to renew the key, e.g. because it crashed or its event loop became too busy, the follower that first sets the key on Redis becomes the new leader, taking over the former leader's tasks. If the previous leader becomes responsive again, it becomes a follower.
```
krynble, line 196:
```suggestion
In a multi-main setup, all `main` handle the leadership process transparently to users. In case the current leader becomes unavailable, e.g. because it crashed or its event loop became too busy, other followers can take over. If the previous leader becomes responsive again, it becomes a follower.
```

I don't think we need to provide this level of details here. Just a suggestion.

docs/hosting/scaling/queue-mode.md
```diff
@@ -181,36 +181,40 @@ n8n worker --concurrency=5
...
-Then, for each main process:
+To deploy n8n in multi-main setup, ensure: 
 
-- Set the environment variable `N8N_MULTI_MAIN_SETUP_ENABLED` to `true`
-- Set a unique `N8N_PORT`.
+- All `main` processes are running in queue mode and are connected to Postgres and Redis.
+- All `main` and `worker` processes are running the same version of n8n.
+- All `main` and `worker` processes have a valid tenant ID and license key.
+- All `main` processes have set the environment variable `N8N_MULTI_MAIN_SETUP_ENABLED` to `true`.
+- All `main` processes have set a unique `N8N_PORT`.
```
krynble, line 208: This is not necessary, if using traeffik or any sort of application load balancer, all containers can use port 5678 and route properly
```suggestion
```

PR #1893, opened 2024-02-09
fixing npm install

soheilade:
npm install n8n g does not work and lead to several issues on fs and patch-package:

https://community.n8n.io/t/n8n-install/39983
https://github.com/n8n-io/n8n-nodes-starter/issues/45

I found npx works like a charm so better to just use it here too as it is meant for new commers.

### DIFF
--- a/_snippets/integrations/creating-nodes/testing.md
+++ b/_snippets/integrations/creating-nodes/testing.md
@@ -2,7 +2,7 @@ You can test your node as you build it by running it in a local n8n instance.
 
 1. Install n8n using npm:
   ```shell
-  npm install n8n -g
+  npx n8n
   ```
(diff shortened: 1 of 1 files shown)


StarfallProjects:
@Joffcom am I right in thinking this is the same issue I encountered a few days ago, and should be fixed by Adi's PR? https://github.com/n8n-io/n8n/pull/8568 Or is it a separate issue?

Joffcom:
@StarfallProjects you got it, So this should be fixed in the next release and it is the same as the issue you had.

StarfallProjects:
Thanks @Joffcom 

@soheilade thanks for being proactive and contributing 🎉 But as this was due to a bug and should shortly be fixed, I'm closing this.

PR #1896, opened 2024-02-12
test cred-only and core node msgs, plus a numbering oddity fix

freakwriter:
Sample message edit to add messages in one core node (activationtrigger) and one cred-only node (qradar).

For Qradar, there was an existing external reference in the section already, so I combined that with the cred-only wording (wasn't sure if that was correct or I should have placed it separately below it?)

I also noticed a list numbering issue when I first opened the Action Network cred node as a point of comparison (it's not in the list in my task), but figured I would fix it since I saw it anyway.

Let me know what I might have messed up here and then I should be good to start rolling through these next week!


StarfallProjects:
@freakwriter thanks for this:

- Activation Trigger is great
- For credential only nodes please do them as separate paragraphs. One is the link to the product's docs to help people find their API keys etc., and the other is the link to examples, so they're distinct things.
- Thanks for fixing the list. However if you see things in a section titled "Example" it's usually not worth fixing, as that's the section you're going to be deleting soon.

PR #1899, opened 2024-02-13
template library semi-removal

StarfallProjects:
…osters isn't hidden in embed


nik8n:
@StarfallProjects do you think we need something like `How to use a template`? Shouldn't that be self-explanatory? I'm asking because we're currently changing that and it might need another update soon. Personally, I'd say we can live without the section

StarfallProjects:
@nik8n I've updated it to mostly remove the section, but keep the bit that tells them how to get there, and explain the differences between n8n library and custom library.

nik8n (approved):

PR #1902, opened 2024-02-15
add auth

StarfallProjects:
@Joffcom not 100% sure if this is what you actually meant in the ticket? Could people use both `authenticate` and `auth`?


Joffcom:
@StarfallProjects Auth is a property like body / header / qs it lets you use Basic Auth as an authentication option so you would use it like...

```
authenticate: IAuthenticateGeneric = {
  type: 'generic',
  properties: {
    auth: {
      username: '={{$credentials.username}}',
      password: '={{$credentials.password}}',
    },
  },
};
```

Joffcom (approved):

PR #1934, opened 2024-02-26
New Node: Outlook Trigger

Joffcom:
Waiting on https://github.com/n8n-io/n8n/pull/8656


StarfallProjects:
@Joffcom you'll need to add it to the nav as well

PR #1937, opened 2024-02-26
Remove examples section from all trigger nodes

freakwriter:
Removed Examples section from all trigger nodes, also fixed a typo in Netlify trigger.


StarfallProjects:
@freakwriter can you also delete the images used in these examples (after checking they're not used elsewhere - but very unlikely that they are)

StarfallProjects (changes requested):
Please also delete the images used in these examples. They shouldn't be used elsewhere, but double-check.

freakwriter:
@StarfallProjects I searched for each image path and confirmed I had no results found in the rest of the repo before removing each of the images and the now-empty folders they were in.

I also noticed there were some orphaned files from other nodes that had already had the examples section removed, so I tidied those up as well.

StarfallProjects (approved):

PR #1939, opened 2024-02-28
1.31.0 release note

Joffcom:


docs/release-notes.md
```diff
@@ -28,6 +28,27 @@ n8n uses [semantic versioning](https://semver.org/){:target=_blank .external-lin
 * MINOR version when adding functionality in a backward-compatible manner.
 * PATCH version when making backward-compatible bug fixes.
 
+## n8n@1.31.0
+
+View the [commits](https://github.com/n8n-io/n8n/compare/n8n@1.30.1...n8n@1.31.0){:target=_blank .external-link} for this version.<br />
```
netroy, line 33: All diff urls should use tags in the same tree. This url will not produce the expected result.

```suggestion
View the [commits](https://github.com/n8n-io/n8n/compare/n8n@1.30.0...n8n@1.31.0){:target=_blank .external-link} for this version.<br />
```

netroy (changes requested):

netroy (approved):

PR #1983, opened 2024-03-21
add new data transformation functions

StarfallProjects:
This PR:

- Removes string toDate
- Adds the new functions listed here: https://github.com/n8n-io/n8n/pull/8845 **except** the native JS functions, the ones marked with (**), and the ones that only return themselves (e.g. have excluded int.toFloat)

### DIFF
--- a/_yaml/data-functions.yml
+++ b/_yaml/data-functions.yml
@@ -234,11 +234,14 @@ df_number:
     returns: Boolean
     description: Converts a number to a boolean. <code>0</code> converts to <code>false</code>. All other values convert to <code>true</code>.
   - funcName: toDateTime
+    args:
+      - argName: fmt
+        optional: true
+        longName: Format
+        type: String enum
+        description: Can be <code>ms</code> (milliseconds), <code>s</code> (seconds), or <code>excel</code> (Excel 1900). Defaults to milliseconds.
     returns: Date
     description: Converts a number to a <a href="https://docs.n8n.io/code/cookbook/luxon/">Luxon date object</a>.
-  - funcName: toInt
-    returns: Integer
-    description: Convert a number to an integer by rounding up or down. <code>.5</code> and above rounds up.
 df_date:
   - funcName: beginningOf
     args:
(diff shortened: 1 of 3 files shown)


_yaml/data-functions.yml
```diff
@@ -316,6 +343,15 @@ df_date:
         options:  *df_durationUnit
     returns: Date
     description: Adds a given time period to a Date. Returns either a JavaScript Date or Luxon Date, depending on input.
+  - funcName: toDateTime
+    returns: Date
+    description: Converts a JavaScript date to a <a href="https://docs.n8n.io/code/cookbook/luxon/">Luxon date object</a>.
+  - funcName: toFloat
```
elsmr, line 349: `toFloat` and `toInt` on Date should be excluded, because `toMillis` and `toSeconds` exist on Luxon and are more clear. Or maybe toMillis and toSeconds should be mentioned here?

_yaml/data-functions.yml
```diff
@@ -212,6 +230,15 @@ df_number:
         description: How many decimal places to round to.
     returns: Number
     description: Returns the value of a number rounded to the nearest whole number, unless a decimal place is specified.
+  - funcName: toBoolean
+    returns: Boolean
+    description: Converts a number to a boolean. <code>0</code> converts to <code>false</code>. All other values convert to <code>true</code>.
+  - funcName: toDateTime
+    returns: Date
+    description: Converts a number to a <a href="https://docs.n8n.io/code/cookbook/luxon/">Luxon date object</a>.
+  - funcName: toInt
```
elsmr, line 239: Would also exclude this one because `round()`/`floor()`/`ceil()` are more clear

_yaml/data-functions.yml
```diff
@@ -212,6 +230,15 @@ df_number:
         description: How many decimal places to round to.
     returns: Number
     description: Returns the value of a number rounded to the nearest whole number, unless a decimal place is specified.
+  - funcName: toBoolean
+    returns: Boolean
+    description: Converts a number to a boolean. <code>0</code> converts to <code>false</code>. All other values convert to <code>true</code>.
+  - funcName: toDateTime
+    returns: Date
+    description: Converts a number to a <a href="https://docs.n8n.io/code/cookbook/luxon/">Luxon date object</a>.
```
elsmr, line 238: Could use some detail on the argument, from autocomplete description:

"Converts a number to a DateTime. Defaults to milliseconds. Format can be 'ms' (milliseconds), 's' (seconds) or 'excel' (Excel 1900 format)."

elsmr (approved):

PR #1996, opened 2024-04-01
DOC-787: Course Level 2 style and clarity

freakwriter:
Still finishing this up but I wanted to push a PR to get feedback on some things:

In Chapter 1:
I think people might be dropping from this course quickly because the first lessons here require you to write JS more or less from scratch with very little guidance on how to do it. I don't think we should be teaching folks how to write JS here, but I do wonder if we should just include some copy/paste-able scripts so that people don't get scared off from doing the exercises (which seem focused on getting you familiar with using the Code node). I didn't make major edits here because I wanted to get feedback on that approach--lmk what you think.

In Chapter 2:
- This section's instructions for the exercises felt vague at times, which sort of forced me to look at the solutions before I could try to build them. I reworked this a bit to set explicit steps with some guidance on how to configure them, as well as updating the exercises for current UI options. I wanted to be sure this approach felt correct.
- I updated the Date example from its previous incarnation of using Calculate and adding 5 days to using a Round Up to End of Month in the calculation instead, since 5 days could have bumped things from the end of one month into the first week of the next. However, since this was one of the few places where I made a distinct change to configuration, I figured someone more familiar with the Date & Time node should review it, since there might be an even better way to handle this. Let me know if you think a different approach should be used.

And in Chapter 5:
We have a very brief appearance from Nathan to start this out, but in starting to work through the different subsections, it was a little unclear to me where some of the requirements were coming from. I suspect we should include more conversations with Nathan at the start of each of these sections so that the requirements do make more sense (more aligned with what the first course does); I just wanted to be sure that was an okay approach before I started doing it.

I have a little work in Chapter 5 and 6 remaining so this PR is not yet complete, mainly just wanted to share it to get feedback on some of the points above so I could put finishing touches on at the same time.


freakwriter:
@StarfallProjects I've finished out the tweaks to Chapters 5 and 6 that I could make without Bart's input.

freakwriter:
In working through the quiz for course 2, I felt like there were a few questions that could use updating or that confused me:

Question 4: I believe the correct answer should now be "Split Into node" but that's not an option here:
image

Question 5: I believe the node that's listed in the question should be "Convert to File" now:
image

Question 9: Caribbean has a typo:
image

Question 11: There is no "Split In Batches" node in Workflow 2 anymore, I think this is supposed to be Loop Over Items node? (Also, none of the answer options here aligned with what I had but I'm guessing that may be my operator error):
image

Joffcom:
@freakwriter I have updated the quiz to correct the issues you found.

freakwriter:
> @freakwriter I have updated the quiz to correct the issues you found.

@Joffcom thank you!

PR #1997, opened 2024-04-02
Change search bar input text color for better visibility against background

ayatnkw:
### Motivation 
Search bar input text color is white when search bar is collapsed (default styling from the material theme) and not visible against the white background. I find this to be confusing since users are unable to see any pre-existing text within the search bar. 

CleanShot 2024-04-02 at 09 07 32
### Changes
Change the text color from white to --color-text-dark (#555555)


StarfallProjects (approved):

StarfallProjects:
Good catch!

PR #2041, opened 2024-04-15
add diagram and tidy up the doc a bit

StarfallProjects:


docs/hosting/scaling/queue-mode.md
```diff
@@ -119,7 +127,7 @@ You can view running workers and their performance metrics in n8n by select **Se
 
 When running n8n with queues, all the production workflow executions get processed by worker processes. This means that even the webhook calls get delegated to the worker processes, which might add some overhead and extra latency. However, the manual workflow executions still use the main process.
 
-Redis is used as the message broker, and the database is used to persist data, so access to both is required. Running a distributed system with this setup over SQLite isn't recommended.
+Redis acts as the message broker, and the database persists data, so access to both is required. Running a distributed system with this setup over SQLite isn't recommended.
```
Joffcom, line 130: I would maybe suggest that this setup wiht SQLite isn't supported rather than saying it isn't recommended.

docs/hosting/scaling/queue-mode.md
```diff
@@ -211,7 +219,7 @@ In a multi-main setup, there are two kinds of `main` processes:
 
 ### Leader designation
 
-In a multi-main setup, all `main` handle the leadership process transparently to users. In case the current leader becomes unavailable, e.g. because it crashed or its event loop became too busy, other followers can take over. If the previous leader becomes responsive again, it becomes a follower.
+In a multi-main setup, all `main` handle the leadership process transparently to users. In case the current leader becomes unavailable, for example because it crashed or its event loop became too busy, other followers can take over. If the previous leader becomes responsive again, it becomes a follower.
```
Joffcom, line 222: All `main` what?

docs/hosting/scaling/queue-mode.md
StarfallProjects, line 130:
```suggestion
Redis acts as the message broker, and the database persists data, so access to both is required. Running a distributed system with this setup over SQLite isn't supported.
```

docs/hosting/scaling/queue-mode.md
StarfallProjects, line 222:
```suggestion
In a multi-main setup, all main instances handle the leadership process transparently to users. In case the current leader becomes unavailable, for example because it crashed or its event loop became too busy, other followers can take over. If the previous leader becomes responsive again, it becomes a follower.
```

Joffcom (approved):

PR #2043, opened 2024-04-15
DOC 777: Add more detail for Shopify credentials

freakwriter:
Summary of changes made:

- Added suggestion to create a custom app and link to Shopify docs into **Prerequisites** section
- Added list of required fields to **Supported authentication methods** with guidance on how those fields map to Shopify's UI
- Added Related Resources links based on the credentials template
- Added a **Quirks** section with one hiccup both I and someone in the community forum ran into (the forum post was how I figured out how to solve it, so it seemed like maybe I wasn't the only one with such an issue)

One thing for review: based on the Slack thread, it did not seem like we needed full step-by-step instructions here. I was able to get the Access Token setup working very quickly with the field mappings I added to the doc. But I wasn't sure if I should have put those mappings somewhere else rather than in the Supported authentication methods section.

### DIFF
--- a/docs/integrations/builtin/credentials/shopify.md
+++ b/docs/integrations/builtin/credentials/shopify.md
@@ -15,17 +15,17 @@ You can use these credentials to authenticate the following nodes with Shopify.
 
 Create a [Shopify](https://shopify.com/){:target=_blank .external-link} account.
 
-Then [create a custom app](https://help.shopify.com/en/manual/apps/app-types/custom-apps){:target=_blank .external-link} to authenticate to. Shopify often defaults to OAuth2 token exchange, so that may be the fastest configuration for initial testing.
+Then [create a custom app](https://help.shopify.com/en/manual/apps/app-types/custom-apps){:target=_blank .external-link} to authenticate to.
 
 ## Supported authentication methods
 
 * [Access Token](https://shopify.dev/docs/apps/auth/access-token-types/admin-app-access-tokens){:target=_blank .external-link}. Requires:
     - Shop Subdomain (remove `.myshopify.com`)
-    - Access Token
-    - APP Secret Key
+    - Access Token: In Shopify UI, the **Admin API Access Token**
+    - APP Secret Key: In Shopify UI, the **API Secret Key**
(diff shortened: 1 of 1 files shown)


docs/integrations/builtin/credentials/shopify.md
```diff
@@ -15,12 +15,31 @@ You can use these credentials to authenticate the following nodes with Shopify.
...
+    - API Key
+    - Password
+    - Shop Subdomain (remove `.myshopify.com`)
 
 ## Related resources
 
 Refer to [Shopify's authentication documentation](https://shopify.dev/docs/apps/auth){:target=_blank .external-link} for more information about the service.
+
+<!-- If this is a credential-only node, add a link to the node page on n8n's website. For example: https://n8n.io/integrations/356-gmail/ -->
+This is a credential-only node. Refer to [Shopify](/integrations/builtin/app-nodes/n8n-nodes-base.shopify/) and [Shopify Trigger](/integrations/builtin/trigger-nodes/n8n-nodes-base.shopifytrigger/) to learn more. View [example Shopify workflows and related content](https://n8n.io/integrations/shopify/){:target=_blank .external-link} and [example Shopify Trigger workflows and related content](https://n8n.io/integrations/shopify-trigger/){:target=_blank .external-link} on n8n's website.
```
StarfallProjects, line 40: This isn't a credential-only node. Credential only nodes are ones where we have creds but no app or trigger node.

docs/integrations/builtin/credentials/shopify.md
```diff
@@ -15,12 +15,31 @@ You can use these credentials to authenticate the following nodes with Shopify.
...
+
+## Quirks
```
StarfallProjects, line 42: Can we give meaningful titles for any quirks? So each one would have a title the user could see and jump to easily (as opposed to a single quirks bucket)

StarfallProjects (changes requested):
Comments inline

freakwriter:
Updated per your feedback, @StarfallProjects

StarfallProjects (approved):

PR #2059, opened 2024-04-24
coda.io instead of coda.com

mariaremote:
found a broken link, and fixed it! (they have changed domain)
Thanks :)

### DIFF
--- a/docs/integrations/builtin/credentials/coda.md
+++ b/docs/integrations/builtin/credentials/coda.md
@@ -12,7 +12,7 @@ You can use these credentials to authenticate the following nodes with Coda.
 
 ## Prerequisites
 
-Create a [Coda](https://www.coda.io/) account.
+Create a [Coda](https://www.coda.io/){:target=_blank .external-link} account.
 
 ## Using Access Token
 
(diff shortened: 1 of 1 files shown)


docs/integrations/builtin/credentials/coda.md
```diff
@@ -12,7 +12,7 @@ You can use these credentials to authenticate the following nodes with Coda.
 
 ## Prerequisites
 
-Create a [Coda](https://www.coda.com/) account.
+Create a [Coda](https://www.coda.io/) account.
```
StarfallProjects, line 15:
```suggestion
Create a [Coda](https://www.coda.io/){:target=_blank .external-link} account.
```

StarfallProjects: Open external links in new tab. I realise it didn't have this before - this is something we're fixing as we see instances of it.

StarfallProjects (changes requested):
Link in new tab (see suggestion)

docs/integrations/builtin/credentials/coda.md
mariaremote: Oh yeah you are totally right! I didn't think of adding it, but will now remember for sure 😉

StarfallProjects (approved):

PR #2075, opened 2024-04-30
Add note about GitHub private access tokens

gurisko:


StarfallProjects:
@Joffcom should we update the creds modal here as well?

PR #2098, opened 2024-05-13
docs: Add Tools agent type docs

OlegIvaniv:
https://github.com/n8n-io/n8n/pull/9339

### DIFF
--- a/docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.agent.md
+++ b/docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.agent.md
@@ -17,7 +17,7 @@ For usage examples and templates to help you get started, refer to n8n's [Agent 
 You must connect at least one tool sub-node.
 ///
 
-n8n provides several agents. The conversational agent is the default. n8n recommends using this for most use cases: it's the easiest to configure and will handle most scenarios. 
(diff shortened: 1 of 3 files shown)


docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.agent.md
```diff
@@ -162,6 +170,38 @@ Use the options to refine the agent's behavior.
 
 You can view prompt examples in the node.
 
+## Tools Agent 
+
+Use the Tools Agent node to use [Langchain's tool calling](https://js.langchain.com/docs/modules/agents/agent_types/tool_calling){:target=_blank .external-link} interface. These are models that detect when a function should be called and respond with the inputs that should be passed to the function. Additionally, this agent has a improved output parsing capabilities since the parser will be passed as a formatting tool.
```
StarfallProjects, line 175:
```suggestion
Use the Tools Agent node to use [Langchain's tool calling](https://js.langchain.com/docs/modules/agents/agent_types/tool_calling){:target=_blank .external-link} interface. These are models that detect when a function should be called and respond with the inputs that should be passed to the function. Additionally, this agent has improved output parsing capabilities as the parser will be passed as a formatting tool.
```

StarfallProjects (commented):
This looks good, but the dev work still looks early days/draft? So leaving this as in review until a bit closer to release when we can do a final check for updates.

PR #2100, opened 2024-05-13
DOC-861: H-L

freakwriter:
Credentials nodes updates for nodes beginning with the letters H through L, except for:

- HTTP Request
- LinkedIn


freakwriter:
@StarfallProjects I'm still working on this one, but could you review the HubSpot credential? I made some major changes to that and I'd like to be sure I am on the correct path. In a nutshell:

The Developer API auth method wasn't previously documented at all.

After some research, I surfaced two issues in Linear that make it seem like the Developer API is required for the Trigger node:
- [N8N-1859](https://linear.app/n8n/issue/N8N-1859/credential-documentation-for-hubspot-developer-api)
- [NODE-342](https://linear.app/n8n/issue/NODE-342/hubspot-trigger-credential-mark-developer-api-key-as-required)

So I added the Developer API instructions, tested them as I went to be sure they worked, and added some text indicating which authentication methods were compatible with which nodes.

I also split out the section on recommended scopes into their own sections so they could be linked to.

All of that is a bit outside of the standard template. I'm guessing you might have more feedback on this set of changes than others, so I was hoping I could get you to review only that file right now.

docs/integrations/builtin/credentials/hubspot.md
```diff
@@ -6,70 +6,89 @@ contentType: integration
...
 
-Create a [HubSpot](https://www.hubspot.com/){:target=_blank .external-link} account.
+Create a [HubSpot](https://www.hubspot.com/){:target=_blank .external-link} account or [HubSpot developer](https://developers.hubspot.com/){:target=_blank .external-link} account.
 
-## Using OAuth
+## Supported authentication methods
 
-/// note | Note for n8n Cloud users
-You can skip these steps. Enter the credential name, then select **Connect my account** in the OAuth section to connect your HubSpot account to n8n.
+- API App token: Used with [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node
```
StarfallProjects, line 20:
```suggestion
- API app token: Used with [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node
```

docs/integrations/builtin/credentials/hubspot.md
```diff
@@ -6,70 +6,89 @@ contentType: integration
...
+
+## Using OAuth2
+
+--8<-- "_snippets/integrations/builtin/credentials/cloud-oauth-button.md"
+
+If you need to configure OAuth2 from scratch, you'll need to create a public app. Refer to the instructions in the [HubSpot Public Apps documentation](https://developers.hubspot.com/docs/api/creating-an-app){:target=_blank .external-link}. If you need more detail on what's happening in the OAuth web flow, refer to the [HubSpot Working with OAuth documentation](https://developers.hubspot.com/docs/api/working-with-oauth){:target=_blank .external-link}.
+
+Authenticating the credential can fail if you don't use appropriate scopes. See [Recommended scopes for HubSpot node](#recommended-scopes-for-hubspot-node) and/or [Recommended scopes for HubSpot Trigger node](#recommended-scopes-for-hubspot-trigger-node) for suggested scopes for this app.
+
+## Recommended scopes for HubSpot node
```
StarfallProjects, line 58: I don't think these are just recommended, I think they're required. Can you rephrase at appropriate points to make clear they need to grant these scopes.

Are these just for OAuth, or are they for API App Token as well? If they're just for OAuth, I would make them heading level 3 so they nest within the OAuth section.

docs/integrations/builtin/credentials/hubspot.md
StarfallProjects: Should this be API app token, or just app token? In the interface it's just app token, and I wonder if calling it API app token could confuse users with the old deprecated API key credential?

freakwriter: Oh that's an excellent idea for clarity!

docs/integrations/builtin/credentials/hubspot.md
freakwriter: Good call. Adjustments made:
- All scopes shifted from recommended/suggested to required
- Added required scopes to the App token situation (original instructions didn't include these, but you're right--they'd both need it)
- Removed the required scopes for trigger node references from OAuth2 (and the App token)
- Since the trigger node really only works with the Developer API auth method, I shifted the required scopes for trigger node to a heading level 3 and moved it within the Developer API method. I left the Hubspot node required scopes as heading level 2 since it's now referenced by both App token and OAuth2.

freakwriter:
@StarfallProjects two of these credentials have a much lengthier title in the ToC:
- LDAP credentials - n8n Documentation
- LoneScale credentials - n8n Documentation

I was guessing for LDAP this might have been done for SEO purposes, but I couldn't really think of a reason why it was done for LoneScale.

Do you know if there's a rationale here? I'm wondering if either one was an accident or, if there was an intention for it, my suggestion would be to add a comment in each file to explain why the lengthier title is being used.

StarfallProjects:
No reason for those titles - have fixed.

docs/integrations/builtin/credentials/lingvanex.md
```diff
@@ -6,28 +6,27 @@ contentType: integration
...
-</div>
+Refer to [Lingvanex's Cloud API documentation](https://docs.lingvanex.com/reference/overview){:target=_blank .external-link} for more information about the service.
+
+## Using API key
+
+To configure this credential, you'll need:
+
+- An **API Key**: Generate an API key from your **Account** page. Refer to [Where can I get the authorization key?](https://docs.lingvanex.com/reference/translator-service-faq#where-can-i-get-the-authorization-key){:target=_blank .external-link} for more detailed instructions.
+
+You'll [receive a $1 credit when you create an API key](https://docs.lingvanex.com/reference/translator-service-faq#is-there-a-free-plan){:target=_blank .external-link}.
```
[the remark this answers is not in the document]
StarfallProjects: This feels like a detail that could easily change. Do you think there is a strong reason to include it?

StarfallProjects (changes requested):
Just one detail I want to confirm if we need to include - see inline comments

docs/integrations/builtin/credentials/lingvanex.md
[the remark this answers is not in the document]
freakwriter: The original write-up had more information about this, so this was me condensing it down. But I'll happily remove it, since it seems like an odd thing for us to reference.

freakwriter:
> Just one detail I want to confirm if we need to include - see inline comments

I accepted that suggestion, since I was also on the fence about that detail. I think this is good to go now then?

PR #2122, opened 2024-05-23
update diagram and add the project tables to the list

StarfallProjects:


krynble (approved):
Looking good.

Even though I don't think we have foreign keys, the `webhook_entity`'s `workflowId` column relates to the `workflow_entity` table.

Also the `role` table is currently not being used, we'll change and use it when we implement the custom roles in RBAC.

PR #2130, opened 2024-05-28
DOC-795: Additional creds guidance for some heavily used creds pages

freakwriter:
So far, this PR includes:

- Google creds (most of which were very detailed, I made a few tweaks and built out the Google AI cred)
- HTTP Request cred
- Ollama cred

### DIFF
--- a/docs/integrations/builtin/credentials/google/oauth-generic.md
+++ b/docs/integrations/builtin/credentials/google/oauth-generic.md
@@ -22,7 +22,14 @@ This document contains instructions for creating a generic OAuth2 Google credent
 
 ### Create a new credential in n8n
 
-1. Follow the steps to [Create a credential](/credentials/add-edit-credentials/). If you create a credential by selecting **Create new** in the credentials dropdown in a node, n8n automatically creates the correct credential type for that node. If you select **Credentials > New**, you must browse for the credential type. To create a credential for a [custom API call](/integrations/custom-operations/), select **Google OAuth2 API**. This allows you to create a generic credential, then set its scopes.
+1. Follow the steps to [Create a credential](/credentials/add-edit-credentials/). 
+
+    /// note | Generic and specific credentials
+    If you create a credential by selecting **Create new** in the credentials dropdown in a node, n8n automatically creates the correct credential type for that node. If you select **Credentials > New**, you must browse for the credential type:
+
+	* To create a credential for a [custom API call](/integrations/custom-operations/), select **Google OAuth2 API**. This allows you to create a generic credential, then set its scopes.
+    ///
+
 2. Note the **OAuth Redirect URL** from the node credential modal. You'll need this in the next section.
 
 	??? Details "View screenshot"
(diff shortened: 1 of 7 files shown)


freakwriter:
@StarfallProjects this is a WIP, but since I'll be working Thursday this week, I was hoping I could get an early review on the HTTP Request cred.

I restructured this one a bit, tried to add more detail for the generic auth types, and added a lot more detail on the SSL cert section which seemed very bare bones. But given how heavily used this cred and node are, I'm curious what your initial thoughts are. I wasn't sure if the different OAuth2 Grant Types might be a good argument for collapsing sections, given how long and rather unwieldy the page has now become? I'm just wondering if there might be a better way to format the whole page than what I've done here...

docs/integrations/builtin/credentials/google/oauth-generic.md
```diff
@@ -22,7 +22,14 @@ This document contains instructions for creating a generic OAuth2 Google credent
 
 ### Create a new credential in n8n
 
-1. Follow the steps to [Create a credential](/credentials/add-edit-credentials/). If you create a credential by selecting **Create new** in the credentials dropdown in a node, n8n automatically creates the correct credential type for that node. If you select **Credentials > New**, you must browse for the credential type. To create a credential for a [custom API call](/integrations/custom-operations/), select **Google OAuth2 API**. This allows you to create a generic credential, then set its scopes.
+1. Follow the steps to [Create a credential](/credentials/add-edit-credentials/). 
+
+    /// note | Generic and specific credentials
+    If you create a credential by selecting **Create new** in the credentials dropdown in a node, n8n automatically creates the correct credential type for that node. If you select **Credentials > New**, you must browse for the credential type:
+
+	* To create a credential for a [custom API call](/integrations/custom-operations/), select **Google OAuth2 API**. This allows you to create a generic credential, then set its scopes.
```
StarfallProjects, line 30: Don't think this needs to be a bullet point?

docs/integrations/builtin/credentials/httprequest.md
```diff
@@ -14,76 +14,164 @@ You can use these credentials to authenticate the following nodes:
 
 You must use the authentication method required by the app or service you want to query.
 
-### Predefined credential types
+## Supported authentication methods
```
StarfallProjects, line 17: If I read 'supported auth type' I expect the content that's in generic (all the different HTTP auth types) I would rename this something like 'Generic and predefined credentials', and include a sentence or two explaining what they are, perhaps with link to other relevant docs?

docs/integrations/builtin/credentials/httprequest.md
```diff
@@ -14,76 +14,164 @@ You can use these credentials to authenticate the following nodes:
...
-    * **Access Token URL**
-    * **Client ID**
-    * **Client Secret**
-    * **Scope**
-    * **Auth URI Query Parameters**
-    * **Authentication**
-3. Select **Save** to save your credentials.
+Use this generic authentication if the app or service you want to use supports OAuth2 authentication.
+
+Requirements to configure this credential depend on the **Grant Type** selected. Refer to [OAuth Grant Types](https://oauth.net/2/grant-types/) for more information on each grant type:
```
StarfallProjects, line 101:
```suggestion
Requirements to configure this credential depend on the **Grant Type** selected. Refer to [OAuth Grant Types](https://oauth.net/2/grant-types/) for more information on each grant type.
```

StarfallProjects (changes requested):
In addition to my comments, can you take a look at the issues flagged by Vale? I'd certainly avoid 'via', and squish passive voice as much as possible

docs/integrations/builtin/credentials/google/oauth-generic.md
freakwriter: Would you just want the second point to remain a bullet point, but not the first?

PR #2133, opened 2024-05-30
DOC-887: Additional creds guidance for Google, Ollama, HTTP Request

freakwriter:
Summary of changes:
- Google creds: Built out the Google AI cred; the rest were very detailed, so those I just made a few tweaks to
- HTTP Request cred: Much more detailed overhaul of the available credential options plus adding SSL cert. The page is fairly long, so I'm open to ways to handle that more elegantly
- Ollama cred: Had no detail previously; added a bit, but this is a fairly simple credential to set up so there's still not a lot.

### DIFF
--- a/_snippets/integrations/managed-google-oauth.md
+++ b/_snippets/integrations/managed-google-oauth.md
@@ -1,5 +1,5 @@
 /// note | Note for n8n Cloud users
-For the following nodes, you can authenticate by selecting **Sign in with Google** in the OAuth section. 
+For the following nodes, you can authenticate by selecting **Sign in with Google** in the OAuth section: 
 
 * [Google Calendar](/integrations/builtin/app-nodes/n8n-nodes-base.googlecalendar/)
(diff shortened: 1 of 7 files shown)


freakwriter:
@StarfallProjects I've reworked HTTP Request based on our conversation, though I'm still on the fence about how long it is.

There was one comment of yours in the old PR I wasn't quite sure what to do with, figured we could discuss next week: https://github.com/n8n-io/n8n-docs/pull/2130#discussion_r1618417716

I added to that but wasn't sure how to move it over here since it's a new PR...

StarfallProjects:
@freakwriter re. that comment: I wouldn't generally put a bullet point for a single item, only if there were two or more (to make a list)

freakwriter:
@StarfallProjects ah, I just hadn't looked at the OAuth generic closely enough, I didn't realize it only had one bullet point. I've removed the orphaned bullet point and just added a paragraph break within the note. Let me know if there are other changes you'd like to see.

StarfallProjects (approved):

PR #2165, opened 2024-06-13
Doc 790 memory

StarfallProjects:
Add the memory concept page.
No example for this one.


docs/_includes/integrations/cluster-nodes/memory-shared.html
```diff
@@ -0,0 +1 @@
+If you add more than one [[ page.meta.title ]] node to your workflow, all nodes access the same memory instance. Be careful when doing destructive actions that override existing memory contents, such as the override all messages operation in the Chat Memory Manager node.
```
OlegIvaniv, line 1: Perhaps we could specify here that the accessed instance depends on `sessionId`, so you definitely could have more memory instances(with different `seesionId`'s in the same workflow

OlegIvaniv (approved):

PR #2171, opened 2024-06-17
DOC-842: Overhaul remaining complex credentials

freakwriter:
Updates these credentials:

- AMQP and FTP (delayed until I had a better feel for how these general/non-provider-specific credentials work)
- Filemaker: From when I didn't know how to check the codebase to verify which API was being called. #themoreyouknow
- LinkedIn: Needed to sift through the differences between the Auth methods and setup a bit more
- Generic S3 credential + app node
- Automizy: Updated to template even though this service no longer seems to exist
- Autopilot: Updated to template but included statement that this wouldn't work with Ortto, added Vale exception for Ortto
- Citrix ADC: Had to figure out how to handle the rebrand
- Magento 2: Updated to template the best I could
- Philips Hue: Made the decision to tell people they didn't need the App ID if they used built-in Cloud OAuth (because you don't) and updated rest to template
- Segment: Updated to template since our existing instructions do still seem to work, even though I'm not sure the app node does.


freakwriter:
@Joffcom would you be able to review the changes to `docs/integrations/builtin/credentials/linkedin.md`? I *think* I've captured the broad strokes of your comments in [this forum thread](https://community.n8n.io/t/linkedin-credential-errors/29363/32) and [this PR](https://github.com/n8n-io/n8n/pull/7451) but this still feels like I could be providing better guidance/more detail about when to use each auth type and other tricks/gotchas about setting them up. I'd appreciate any insight or feedback you might have on how to make this better!

freakwriter:
@StarfallProjects This includes anything that I had marked as more complex and needed to return to, as well as the creds I initially skipped due to complexity or issues. I've at least updated everything to template and tried to add statements about known limitations, etc. This one might be a trickier review but it's the last of the initial credentials updates!

StarfallProjects (approved):

PR #2179, opened 2024-06-24
DOC-759: Loop Over Items docs updates

freakwriter:
Updated:

- Adjust .context syntax to current syntax
- Tweaks for clarity on what happens when the node is "done"

Still to do:

- Add clarity on how reset works and how to use the .context commands.

### DIFF
--- a/docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
+++ b/docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
@@ -64,9 +64,9 @@ The final workflow looks like this:
 
 ## Check that the node has processed all items
 
-To check if the node still has items to process, use the following expression: `{{$("Loop Over Items").context["noItemsLeft"]}}`. This expression returns a boolean value. If the node still has data to process, the expression returns `false`, otherwise it returns `true`.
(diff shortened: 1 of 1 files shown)


docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
```diff
@@ -8,14 +8,15 @@ contentType: integration
 
 The Loop Over Items node helps you loop through data.
 
-The node saves the original incoming data, and with each iteration, returns a predefined amount of data through the **loop** output.
+The node saves the original incoming data, and with each iteration, returns a predefined amount of data through the **Loop** output.
 
-When the node execution completes, it combines all the data and returns it through the **done** output.
+The Loop Over Items node stops executing after all the incoming items get divided into batches and passed on to the next node in the workflow, so it's not necessary to add an IF node to stop the loop. When the node execution completes, it combines all the data and returns it through the **Done** output.
```
StarfallProjects, line 13:
```suggestion
The Loop Over Items node stops executing after all the incoming items get divided into batches and passed on to the next node in the workflow, so it's not necessary to add an IF node to stop the loop. When the node execution completes, it combines all the data and returns it through the **done** output. 
```

StarfallProjects: Matching the case of UI copy exactly (I know it looks weird, but it's 'done' in the UI)

PR #2195, opened 2024-06-29
fix n8n data directory path

KShivendu:
Hello, thanks for building a great product. 

Looks like you recently change the directory to store n8n files in the container. I lost valuable data because of mounting to wrong path and assuming it would be right all along (as per docs). Took me a while to figure out the right path. I don't want it to happen with anyone else =)

### DIFF
--- a/docs/hosting/installation/docker.md
+++ b/docs/hosting/installation/docker.md
@@ -29,7 +29,7 @@ From your terminal, run:
 ```sh
 docker volume create n8n_data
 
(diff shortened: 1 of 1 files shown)


Joffcom:
Hey @KShivendu,

Sadly this PR would be asking users to use the old directory, since the release of 1.0 the path in the container is `/home/node/.n8n` using `/root/.n8n` is no longer recommended.

You can find the details on this change in the 1.0 migration guide here: https://docs.n8n.io/1-0-migration-checklist/#breaking-changes

PR #2198, opened 2024-07-01
update release note v1.48.2

ayatnkw:


StarfallProjects:
@ayatnkw looks good! However we put releases in date order (meaning if 1.47.2 is later than 1.48.0, it would come higher than it)

StarfallProjects (changes requested):
see comment

StarfallProjects (approved):

PR #2218, opened 2024-07-08
DOC-921: Kate's audit of core nodes

freakwriter:
Made these changes:

- Updated core nodes to current template
- Added widget template for Templates & examples section, removed other notes/links to those examples and templates
- Ensured all parameters and options are documented.


freakwriter:
@StarfallProjects I think this is finally ready for review, minus the handful of issues with the macro I've added to the spreadsheet.

One change of note:
I've added a new snippet for data comparison nodes like If, Filter, etc., that will list the different data types and the comparisons they support, and I added that into the core-nodes template to try to self-document that it exists. Hopefully that was the right decision here.

docs/integrations/builtin/core-nodes/n8n-nodes-base.manualworkflowtrigger.md
```diff
@@ -11,6 +11,7 @@ Use this node if you want to start a workflow by selecting **Test Workflow**, an
 
 Workflows always need a trigger (start point). In most cases, a trigger node starts the workflow in response to an external event. However, you still need a trigger node even when starting the workflow manually.
 
-/// note | Examples and templates
-For usage examples and templates to help you get started, refer to n8n's [Manual Trigger integrations](https://n8n.io/integrations/manual-trigger/){:target=_blank .external-link} page.
-///
+## Templates and examples
```
StarfallProjects, line 14: Suggest removing the examples here. This node doesn't really need it, and the templates will just end up being the most popular overall.

StarfallProjects (approved):

PR #2222, opened 2024-07-10
update test page

Joffcom:
Adds widgets to the test page.


StarfallProjects:
While you're at it could you add a test for the glossary? 
It should be like this: https://squidfunk.github.io/mkdocs-material/reference/tooltips/?h=glossary#adding-a-glossary

But Oleg has just encountered problems and submitted a fix: https://github.com/n8n-io/n8n-docs/pull/2223

I am really not sure why it would suddenly break, we've been using it since last year

docs-site-feature-tests/index.md
```diff
@@ -52,3 +52,10 @@ Other images like this should expand on click:
 [This link should show a preview of the quickstart on hover](/try-it-out/quickstart/){ data-preview }
 
 [This link should NOT show a preview on hover](/try-it-out/quickstart/)
+
+## Embed Workflow
```
StarfallProjects, line 56:
```suggestion
## Embed Workflow

You should see an interactive workflow, similar to the templates pages or forum.
```

docs-site-feature-tests/index.md
```diff
@@ -52,3 +52,10 @@ Other images like this should expand on click:
...
+[[ workflowDemo("https://api.n8n.io/workflows/templates/655") ]]
+
+## Templates Widget
```
StarfallProjects, line 59:
```suggestion
## Templates Widget

You should see a list of three templates, followed by links to the integrations page on the website, and the templates search.
```

docs-site-feature-tests/index.md
```diff
@@ -52,3 +52,10 @@ Other images like this should expand on click:
...
+<!-- see https://www.notion.so/n8n/Pull-in-templates-for-the-integrations-pages-37c716837b804d30a33b47475f6e3780 -->
+[[ templatesWidget('Test Title', 'http-request') ]]
```
StarfallProjects, line 61: This won't work. The title of the page is used to create the API call to load the template. You need to provide an actual title if you want it to load templates.

StarfallProjects (changes requested):
See comments

docs-site-feature-tests/index.md
Joffcom: Changed

PR #2231, opened 2024-07-11
DOC-567: Create new Securing n8n section, add placeholder content, update nav

freakwriter:
Includes these changes:

- Created new "Securing n8n" section in /hosting/
- Created rough drafts of content mentioned in Linear [DOC-567](https://linear.app/n8n/issue/DOC-567/securing)
- Turned SAML intro into a snippet so it could be reused here and in user-management
- Updated nav to include new pages
- Moved security audit from monitoring & logging to here, but still kept it in the nav for Monitoring & logging


freakwriter:
Things I need help on:

- Is there any other content that should go in here?
- Is the content in here actually correct? I was particularly unsure on SSL. Need review by at least one SME.

freakwriter:
@Joffcom and @csuermann: Deb transitioned this one to me before she left and mentioned that you two might be good SMEs to review the content. I pulled together some existing content I could find, but I suspect one or both of you likely have way more you could add. Would you be able to take a look and let me know here or in Slack what else you think needs to be included and/or what info is inaccurate?

docs/hosting/securing/set-up-ssl.md
```diff
@@ -0,0 +1,25 @@
+---
+#https://www.notion.so/n8n/Frontmatter-432c2b8dff1f43d4b1c8d20075510fe4
+description: "Set up SSL for your self-hosted n8n instance."
+contentType: howto
+---
+
+# Set up SSL
```
csuermann, line 7: The preferred way of setting up TLS is by using a reverse proxy like Traefik which also takes care of certificate renewals.

netroy: I agree. We could maybe create 2 sections here:
1. using a reverse proxy to outsource TLS management on ingress, or
2. passing in certs to n8n directly

We should then mark 1 as "Recommended", and 2 as "Not Recommended".

I'd also be fine with us skipping 2, and deprecating that option in one of the future releases.

Joffcom: I would add both options to the docs just to be complete, I am also not sure on removing the option from the application if anything we should be creating certs by default given that we don't allow http outside of localhost anymore.

We should still make sure that using a reverse proxy or NLB is marked clearly as preferred.

freakwriter: Thanks for the feedback, everyone.

freakwriter:
@ayatnkw could you review for style/tone and any other changes you see fit?

ayatnkw (approved):
looks great!

PR #2232, opened 2024-07-12
node tweaks

Joffcom:
Renames Citrix <> Netscaler
Adds redirects


docs/integrations/builtin/app-nodes/n8n-nodes-base.netscaleradc.md
```diff
@@ -0,0 +1,38 @@
...
+	* Install
+* File
+	* Delete
+	* Download
+	* Upload
+
+## Templates and examples
+
+<!-- see https://www.notion.so/n8n/Pull-in-templates-for-the-integrations-pages-37c716837b804d30a33b47475f6e3780 -->
+[[ templatesWidget(title, 'netscaler-adc') ]]
```
freakwriter, line 31: @Joffcom it looks like the marketing website still has "Citrix ADC", so the link this generates throws a 404. Deb and I had resolved some of these by adjusting the frontmatter title to match the marketing site, but I'm hesitant to do that here. Do you know if the marketing site will also be updated, so this would work as expected?

freakwriter:
@Joffcom it looked like Netscaler was throwing some Vale issues, so I updated accept.txt to allow that.

See in-line file comments for other feedback.

One follow-up question: Should the Cal Trigger docs also be updated to Cal.com?

docs/integrations/builtin/app-nodes/n8n-nodes-base.netscaleradc.md
Joffcom: @freakwriter That is a good point, I would expect the changes to be made on the marketing site as well. We will need to let them know.

Joffcom:
@freakwriter I thought about the rename but I am not sure how mkdocs would handle that sort of naming.

freakwriter (approved):

PR #2247, opened 2024-07-16
DOC-958: Add priority frontmatter to all medium, high, and critical nodes and related credentials

freakwriter:
Ignoring "low" priority since that's implied by no priority.


freakwriter:
@ayatnkw does this format for the front-matter work for you? I've added `priority: critical`, `priority: high`, and `priority: medium` into the front-matter, below the normal text strings like title, description, and contentType, but above any lists of tags in the front-matter.

I figure anything without a priority we'll assume is low. :)

ayatnkw (approved):
I haven't checked if all the nodes have the correct priority for now, as I think we can do this as we update each node. But overall, looks good!

PR #2249, opened 2024-07-17
Azure external secrets

Joffcom:


Johannesishere (approved):
I would like the explanation to be more detailed. Will approve now, since I need to link to this for the release notes.

PR #2255, opened 2024-07-22
DOC-959: Update Telegram credentials, node, and trigger for Critical node priority

freakwriter:
This PR includes these updates:

- Adds more detail to Telegram credential page, per Critical priority
- Expands the Telegram node to include all parameters and additional fields/options, add links to relevant docs and method for each resource + action combo, other wording and structural changes
- Update the Telegram trigger page to fit current template


docs/integrations/builtin/trigger-nodes/n8n-nodes-base.telegramtrigger.md
```diff
@@ -18,15 +18,23 @@ You can find authentication information for this node [here](/integrations/built
...
+## Events
+
+- **`*`**: Triggers on all updates of any kind.
+- **Callback Query**: Triggers on a new incoming callback query.
+- **Channel Post**: Triggers on a new incoming channel post of any kind, including text, photo, sticker, and so on.
+- **Edited Channel Post**: Triggers on a new version of a channel post that's known to the bot and was edited.
+- **Edited Message**: Triggers on a new version of a message that's known to the bot and was edited.
+- **Inline Query**: Triggers on a new incoming inline query.
+- **Message**: Triggers on a new incoming message of any kind, including text, photo, sticker, and so on.
+- **Poll Change**: Triggers when the poll state changes. Bots receive updates about stopped polls and polls which are sent by the bot.
```
ayatnkw, line 30: In the UI, the option is called 'Poll', is there a reason why you've renamed it this way?

docs/integrations/builtin/app-nodes/n8n-nodes-base.telegram.md
```diff
@@ -18,33 +18,629 @@ Refer to [Telegram credentials](/integrations/builtin/credentials/telegram/) for
...
-    * Send Chat Action
-    * Send Document
-    * Send Location
-    * Send Media Group
-    * Send Message
-    * Send Photo
-    * Send Sticker
-    * Send Video
-    * Unpin Chat Message
+* Chat operations
```
ayatnkw, line 21:
```suggestion
* **Chat** operations
```
What do you think about making the resource names (Chat, Callback, File, Message) bold since they're also an UI option?

ayatnkw (commented):
Questions:
When there are lots of options, the workflow template section gets pushed to the very bottom making it less visible. Should the template section come before the operations section?

freakwriter:
> Questions: When there are lots of options, the workflow template section gets pushed to the very bottom making it less visible. Should the template section come before the operations section?

This is a good question, since it's one of those questions around "how do we think people are using these docs/where would they get the most value"?

@nik8n I'd be interested in your perspective on this. Take a look at the overhauled [Telegram node](https://deploy-preview-2255--n8n-docs.netlify.app/integrations/builtin/app-nodes/n8n-nodes-base.telegram/).

We're trying to provide a lot more detail on what various operations do/how to configure them, but it means that the "Templates and examples" section (which Deb had me add after Operations) ends up way down the page. Do you think there's value in us bumping the templates and examples above Operations, now that we're having Operations contain a lot more information/guidance? The templates widget itself is pretty new to the docs so I'm not sure we have a lot of info about how often people are using it, but if it feels like it's adding value, moving it above Operations might make more sense, especially as it's going to be a consistent size/length and could be easy for users to skip past if it's not of interest.

nik8n:
I think the structure does not really work like this anyway, does it? I see a two big problems with it right now

1. We double operations (once in the overview, and then the detailed sections about it)
2. It feels like it's WAY too much content. Important things like the templates and also how to get the chat ID are completely lost like this.

Could we not maybe use sub-pages or maybe toggle sections or something? (just thinking out loud)

freakwriter:
@nik8n I've tried a different approach here, adding subpages. This keeps a list of all operations on the main node page, which also has the links to the templates & examples, info about getting the chat ID and other troubleshooting tips, but then each operation type has its own subpage that goes into detailed configuration information. This does seem to improve readability and nav a bit, but it does still leave the Message operations page pretty long. (Maybe this is fine, since there's no content on here that isn't specific to the operations?)

nik8n:
Yeah, I can see how this becomes pretty lengthy. Personally, I feel like this is okay because it meets what I would expect to see. But I guess different users could have different opinions here. Both routes sound like valid options... maybe we saw once a node has more than X operations we do a dedication page for it?

freakwriter:
@nik8n yeah I'm thinking a loose guideline about the number of operations would make sense, and then maybe a judgment call based on how complicated the operations are (some nodes have very simple operations that might not need the separate pages, for example).

The exact number is something we could feel out as we start tackling the app nodes, maybe for now we say more than 6 as an automatic thing and we see how that looks IRL?

nik8n:
Sounds good to me

freakwriter:
@ayatnkw could you review the layout used here? I'm curious what you think about the multiple subpages, etc.

nik8n (approved):
I like the format of it. Maybe the operations are still a little too long. But I feel like this works well.

PR #2257, opened 2024-07-23
update export import page

ayatnkw:


docs/workflows/export-import.md
```diff
@@ -6,14 +6,30 @@ contentType: howto
...
   ![Select a group of nodes](/_images/workflows/export-import/selectingnodes.gif)
 
-You can import JSON files as workflows by copying the JSON workflow to the clipboard (`Ctrl + c` or `cmd +c`) and paste it (`Ctrl + v` or `cmd + v`) into the Editor UI.
+## From the Editor UI menu
+
+From the top navigation bar, select the three dots in the upper right <img alt="Workflow menu icon" class="off-glb" src="/_images/common-icons/three-dots-horizontal.png"> to see the following options: 
+
+<figure><img src="/_images/courses/level-one/chapter-six/l1-c6-import-export-menu.png" alt="Import/Export menu" style="width:100%"><figcaption align = "center"><i>Import & Export workflows menu</i></figcaption></figure>
+* **Download**: Downloads your current workflow as a JSON file to your computer
+* **Import from URL**: Imports a workflow as a published JSON file
```
Joffcom, line 29: I don't think "published" makes sense here, What about:

> Imports workflow JSON from a URL e.g.

docs/workflows/export-import.md
```diff
@@ -6,14 +6,30 @@ contentType: howto
 
 # Export and import workflows
 
+/// warning | Sharing credentials
+Note that the exported JSON file of your workflow will contain your credentials as well. If you don't want to share your credentials (recommended), delete them from the JSON file.
```
Joffcom, line 10: This isn't fully accurate, The workflow JSON won't contain the credential but it will include the name and ID of the credential, The ID is not a sensitive value but the name the user inputs could be.

The bigger risk is potentially HTTP Request nodes that include auth headers in the node itself rather than as generic credentials, This would normally happen if you import from CURL.

Joffcom (changes requested):

docs/workflows/export-import.md
ayatnkw: thanks for clarification, will make changes!

freakwriter (approved):
Looks good to me, I pushed one change to adjust CURL to cURL. I'll defer any other changes to @Joffcom !

Joffcom (approved):

PR #2263, opened 2024-07-23
fix typo in Wait Node Docs #Time-based operations

mariaremote:
just a quick typo fix found in the Wait Node Docs
https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.wait/#time-based-operations

image

- [x] tested with `mkdocs`


freakwriter (approved):
Good catch, @mariaremote . This looks good!

PR #2265, opened 2024-07-23
DOC-977: Update Notion credential to High tier template

freakwriter:
The Notion credential already had detailed step-by-step instructions appropriate for the High tier, so I only made these changes:

- Added public vs. internal integrations section using content from Notion, linked to their doc on it.
- Slight formatting tweaks to align with current template
- Other small wording tweaks for readability


freakwriter:
@nik8n tagging you on this one in case you want to review it before it goes out. It felt like the internal vs. public integration might fall into the "FAQ" category of content and we had no previous guidance on it, so that's the only major change here after changes I released a few weeks ago.

@ayatnkw could you be my style/formatting review 👀?

nik8n (changes requested):
A bit of feedback:
- When mentioning how to share pages with the connection, the section is called `Connections` and the button is called `Connect to` instead of `Add connections` (it seems like the Notion doc is actually out of date here :D)
- I would have an own section for `How to share a page with your integration` instead of hiding it inside the `Using API integration token` section. Otherwise it gets lost very quickly when just skimming over it. WDYT?

All the changes you made make sense to me.

freakwriter:
@nik8n good catch on the Notion docs being incorrect. Derpaderp. And the suggestion about sharing a page with the integration in its own heading makes sense--I hemmed and hawed about that and talked myself out of it, so it's reassuring to have someone else make the same suggestion. I've updated with that feedback and made a couple other small tweaks.

ayatnkw (approved):
Looks good! I love the clarification between internal vs public integrations :) @freakwriter

nik8n (approved):
Awesome, looking good!

PR #2266, opened 2024-07-23
DOC-973: Update WhatsApp Business credential

freakwriter:
I'd already updated this credential a few weeks ago to be more detailed, so this PR just adjusts some small formatting details to better align with the current template and fixes some URLs that are now redirecting.

Updated based on Jon's feedback in #team-nodes that OAuth is for the trigger node and access token is for the app node.


ayatnkw (approved):
Looks good. One question though, I noticed a bit of inconsistency when we refer to a Whatsapp node. (WhatsApp Business Cloud node, WhatsApp Business node, WhatsApp node etc) We use `WhatsApp Business Cloud` in the UI. 

I do think `WhatsApp Business Cloud` is a bit wordy so we can stick to Whatsapp node in the docs? @freakwriter

Joffcom:
Looks good, With the phrasing I think the service is officially called "WhatsApp Business Cloud" but everyone knows it as WhatsApp so it is a bit of a coin flip on if we want to change that.

PR #2268, opened 2024-07-25
Just added Azure to the list at the top of the external secrets page

Johannesishere:


Johannesishere:
Accidentally also added this to my release notes PR. Well better save then sorry, I guess. 😬

PR #2281, opened 2024-07-28
Triggers overview page

alexpeterkonieczny:
Added a page giving an overview of trigger nodes. It contains the following:

1. Definition
2. Add a trigger to your workflow
3. Test a trigger
4. Execute your workflow after adding triggers
5. Types of triggers

This page is a good start. There are a few problems that should be looked into next though.

- I am not sure I have added the file to the correct place within the repository.
- This page lacks images to direct the user. Two particular cases where this would be particulary helpful would be in highlighting the cosemetic differences between how triggers appear in a workflow compared to other nodes and another would be an image to show readers exactly where to go to activate their workflow.
- The example explaining how a schedule trigger would work in production mode is weak as written here. 
- Capitalizations norms need to be solidified when it comes to the different units of a workflow and how they are named. For example, "Trigger" vs "trigger", or even in the case of the latter, "Schedule trigger" vs "schedule trigger". It seems inconsistent to capitalize when node and workflow are lowercase.
- Tone. This page has a Flesch reading ease score of 60.77, which meets the contribution guidelines, but only barely. This came at the sacrifice of some personality in the tone of the writing. Some work could be done here to make the writing not wear on the endurance of the reader too much.

### DIFF
--- a/docs/workflows/components/triggers.md
+++ b/docs/workflows/components/triggers.md
@@ -13,19 +13,19 @@ Triggers have a lot in common with other nodes. However, there are some key diff
 
 For more information, check out the [Nodes](/workflows/components/nodes/) page.
 
-## Adding a trigger to your workflow
+## Add a trigger to your workflow
 
 When you create a new workflow, n8n will automatically add a trigger node. This default node is a manual trigger. Just like in the node panel, you can easily see if a node in your workflow is a trigger. Just check to see if there is a lightning bolt icon next to the node. 
 
 You can add as many triggers as you want. When you activate your workflow, all added triggers will listen for their defined events to know when to start running.
 
-## Testing a trigger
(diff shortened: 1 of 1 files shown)


Joffcom:
Hey @alexpeterkonieczny 

Is this for the technical task? It may be better to keep it in your fork so others are less likely to borrow your work 🙂

alexpeterkonieczny:
@Joffcom Will do. Sorry about that.

alexpeterkonieczny:
@Joffcom are you able to delete the pull request? I don't think I can from my end.

Joffcom:
Hey @alexpeterkonieczny sadly GitHub doesn't allow deleting a pull request (or at least not that I am aware of) but I suspect most people would only look at open rather than closed.

PR #2282, opened 2024-07-29
Quick updates to hosting guides

Joffcom:


freakwriter (approved):
@Joffcom you didn't have a reviewer assigned, but I'm guessing if you had this up you wanted someone to. These changes look good but let me know if you have more coming!

Joffcom:
@freakwriter to be honest it was a throw away idea based on internal feedback but I think it does make sense so I will get it merged. At some point I want to overhaul the guides as there is a lot of duplicate content that could be snippets.

PR #2284, opened 2024-07-29
DOC-968: Google credentials tweaks

freakwriter:
Some wording updates to better align with current template and try to make clearer when you'd want to use which auth method.

This is already highly-detailed.


freakwriter:
@nik8n and @ayatnkw the Google credentials are already highly detailed. I made some tweaks for readability but was curious if y'all had any other feedback on changes you'd like to see here.

@nik8n do you think we should update the YouTube videos here? I haven't touched any of the videos in the docs but it looks like the n8n UI is showing an older version...?

nik8n:
A few things (let's discuss):

1. Remove the `Create a new credential in n8n` section: I don't think we have this anywhere else, so why have it here?
2. Why did you remove the OAUTH consent screen step? That is still needed, isn't it?
3. The video definitely needs an update, but maybe let's do that later / as a separate action? It would make it easier to see the impact.
4. I guess the video should come early (at the top of the page) normally. Maybe we can do that once we do (3)

freakwriter:
@nik8n I didn't remove the OAuth consent screen; I reworded it to make it seem more required and added the note that you can skip it if you've done it before after the required step, so people would be less likely to skip it. It's now this:
* [Configure the OAuth consent screen](https://developers.google.com/workspace/guides/configure-oauth-consent){:target=_blank .external-link}. If you've already used OAuth in your Google Cloud project before, you can skip this step.

Instead of this:
If you haven't used OAuth in your Google Cloud project before, you need to [configure the OAuth consent screen](https://developers.google.com/workspace/guides/configure-oauth-consent){:target=_blank .external-link}.

Since it seems like Google is one of the credentials people want to test with first, I suspect we're better off assuming they haven't configured a consent screen before than assuming they have.

I had wondered about that separate "New credential" thing at the top, too, since these are the only pages I've ever seen it on. I'm assuming it was done at some point in the past to help clarify what people needed to do, but I'm happy to rip it out.I've had some other ideas on ways to improve this, too, so I'll work all those in before I re-request review.

nik8n:
Ahh, it's in the prerequisites now. Are you not afraid that this will get lost? I for example just skip the prerequisites every time. Also in user tests, we see that users are not actually reading the docs, but just skim over it really really quickly.

freakwriter:
It was actually in both the prerequisites and the numbered instructions,
but I'm revamping the instructions because I think the current layout
encourages people to skip over steps. Give me a few more minutes!

ayatnkw (approved):
made some comments in the slack channel (thanks for answering @freakwriter)

But looks good!

freakwriter:
@nik8n Turns out that the OAuth consent screen wasn't properly in the steps for the generic auth, which I think is what you were commenting on (I'd prioritized the single service one and somehow overlooked that the generic auth didn't have identical steps in it). They should be much more clearly laid out now and consistent with each other.

@ayatnkw I both added a screenshot of the dropdown and a link to create a new project directly, as I think they're both good suggestions.

PR #2287, opened 2024-07-30
Quick update for SSO / SAML

Joffcom:
Updates Editor URL env description and adds a note to the SAML setup


ayatnkw (approved):
looks good!

PR #2289, opened 2024-07-30
DOC-978: Update Ollama credential to high tier node template

freakwriter:
This PR mostly just adds a little bit more detail to the auth configuration and adds an FAQ issue from the forums into the page.


freakwriter:
@Joffcom can you review the very bottom of this credential talking about running Ollama in a different container from n8n? It seemed like there were a few forum questions around this and I'm hoping to fend some of those off with this update, but want to be sure it's worded appropriately.

Joffcom:
@freakwriter Looks good to me, I would be tempted to not mention the ngrok thing as I don't think Ollama has any authentication built in so a user exposing it could cause themselves some issues.

ayatnkw (approved):

freakwriter:
@Joffcom thanks, this was why I wanted your eyes on it! Removed that reference and releasing the rest. :)

PR #2290, opened 2024-07-31
Webhook & HTTP Request credentials overhaul

ayatnkw:
Changes made:
- Create a separate Webhook credential page
- Create new docs snippets for each generic_auth type 
- Update both Webhook credential and HTTP Request credential to use those snippets so that we’re consistently describing these generic auth types in both places without copying, rewriting, or cross-referencing content between them


ayatnkw:
@freakwriter I'm going to first work on the app node credentials but wanted to make a draft PR for the changes (wip) I made already in case you wanted to take a look / had feedback on it.

freakwriter (approved):

PR #2292, opened 2024-07-31
update supabase credentials

ayatnkw:
Changes made:
- updated the sign up link
- updated the formatting to numbered list and added more details


freakwriter (approved):
I pushed a small commit to fix one heading that was in title case (which apparently we both missed along the way!) and to turn Vale off in the final steps because it was fully erroring on the service_role section. Good to go now!

PR #2296, opened 2024-08-01
DOC-981: Update Baserow credential to fit high tier format

freakwriter:
Very minimal changes here to align with current template; moved the detailed stuff into a numbered list and compressed the bullet point list slightly since it's so simple.


freakwriter:
@ayatnkw this one should be a very fast review.

ayatnkw (approved):

PR #2301, opened 2024-08-01
DOC-969: Overhaul Send Email credentials

freakwriter:
Summary of changes:

- Split single page into folder with separate page for each email provider
- Change email provider instructions to detailed step-by-step instructions


freakwriter:
@nik8n and @ayatnkw : I'm testing a different layout for Send Email, so that each provider-specific set of instructions is on its own page. I'm thinking this might be a good layout for Send Email and IMAP. Can you take a look and let me know what you think?

ayatnkw (approved):
Fixed the broken note for the App password but otherwise looks good! I also think separating the different providers into separate pages is a nice idea!

nik8n:
I really like the idea!

PR #2315, opened 2024-08-05
DOC-1002: Update Groq credentials to template

freakwriter:


freakwriter:
@ayatnkw super simple changes, so this should be a quick one to review!

PR #2333, opened 2024-08-12
Doc 1021 Node overhaul: Google Sheets

ayatnkw:
### Changes
- Added detailed description for each operation and associated options


ayatnkw:
I played around with the formatting and thought about maybe having the Options section as a table instead, what do you think? @freakwriter @nik8n 
- [Here's](https://deploy-preview-2333--n8n-docs.netlify.app/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/) the original format
- [Here's](https://deploy-preview-2333--n8n-docs.netlify.app/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets-v2/) the options as tables

_snippets/integrations/builtin/app-nodes/googlesheets/node-options-v2.md
```diff
@@ -0,0 +1,8 @@
+#### Options
+
+| **Options** | **Description** |
```
freakwriter, line 3:
```suggestion
| **Option** | **Description** |
```

freakwriter:
@ayatnkw Options are handled somewhat inconsistently across nodes. Things I like about your table approach:

-  The Options section has its own subheading. I'd do this whether you're using a table or not.
- The different options for Recalculation Interval as a separate sub-list rather than stuck in a paragraph of text. I'd similarly use this format whether you're using a table or not, for easier scanning/reading. (I'd also suggest using this format for the options for Locale.)

One of the advantages to the table could be splitting out the description of what the option's used for from the actual options you can choose from and/or having some kind of "Related resources" section that could contain the links to relevant external docs. I might try playing around with that to see if we can make the table do even more than it's currently doing, something like:

| Option | Description | Configuration | Related resources |

To me, that takes it from just a change of presentation to one of really making the most of the format change to convey more information faster, so the format does more for us. Do you want to try that? Should I try adding a suggested commit to show what I'm thinking in more detail?

nik8n:
Wouldn't we move the operation details to a different page here as well? Just like we did with Telegram?

That would probably have a few advantages:

1. The site would be cleaner and easier to scan
2. We could allow options to be a header on the sub-page without cluttering it too much.

In any case, I think I slightly prefer the table since it feels a little cleaner. But I think we should always have sub-pages for the operation details.

freakwriter:
I hadn't looked at the overall length here, but I suspect Nik's suggestion is a good one, to play around with the subpage layout for Document vs. Sheet Within a Document operations...that might give us some more visual space to work with and then having the subheading for Options isn't going to feel as overwhelming. @ayatnkw maybe this would be a good one for us to work together on to play around with that, and then we can see if this number of operations feels like it's close to the cutoff for having the subpage format??

ayatnkw:
Thanks for the comments, I'll put the operation details in separate pages and play around with the table formatting. 

One thing I'm uncertain about with the table formatting is how well the AI assistant will read it. I'm also not so sure about having too many columns because of the limited width on the docs page (I'd say max 3 columns) I'll play around a bit and maybe @freakwriter and I can talk about it on Monday.

ayatnkw:
@nik8n @freakwriter made suggested changes, would be great if you could take another look!

docs/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/sheet-operations.md
```diff
@@ -0,0 +1,220 @@
...
+
+Enter these parameters:
+
+- **Credential to connect with**: Create or select an existing [Google Sheets credentials](/integrations/builtin/credentials/google/).
+- **Resource**: Select **Sheet Within Document**.
+- **Operation**: Select **Delete**.
+- **Document**: Choose a spreadsheet that contains the sheet you want to delete. 
+    - Select **From list** to choose the spreadsheet title from the dropdown list, **By URL** to enter the url of the spreadsheet, or **By ID** to enter the `spreadsheetId`. 
+    - You can find the `spreadsheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/spreadsheetId/edit#gid=0`.
+- **Sheet**: Choose a sheet in which you want to delete.
```
freakwriter, line 126:
```suggestion
- **Sheet**: Choose the sheet you want to delete. 
```

docs/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/sheet-operations.md
```diff
@@ -0,0 +1,220 @@
...
+
+Enter these parameters:
+
+- **Credential to connect with**: Create or select an existing [Google Sheets credentials](/integrations/builtin/credentials/google/).
+- **Resource**: Select **Sheet Within Document**.
+- **Operation**: Select **Delete Rows or Columns**.
+- **Document**: Choose a spreadsheet that contains the sheet you want to delete rows or columns from. 
+    - Select **From list** to choose the spreadsheet title from the dropdown list, **By URL** to enter the url of the spreadsheet, or **By ID** to enter the `spreadsheetId`. 
+    - You can find the `spreadsheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/spreadsheetId/edit#gid=0`.
+- **Sheet**: Choose a sheet in which you want to delete.
```
freakwriter, line 144:
```suggestion
- **Sheet**: Choose the sheet in which you want to delete rows or columns.
```

docs/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/sheet-operations.md
```diff
@@ -0,0 +1,220 @@
...
+- **Credential to connect with**: Create or select an existing [Google Sheets credentials](/integrations/builtin/credentials/google/).
+- **Resource**: Select **Sheet Within Document**.
+- **Operation**: Select **Delete Rows or Columns**.
+- **Document**: Choose a spreadsheet that contains the sheet you want to delete rows or columns from. 
+    - Select **From list** to choose the spreadsheet title from the dropdown list, **By URL** to enter the url of the spreadsheet, or **By ID** to enter the `spreadsheetId`. 
+    - You can find the `spreadsheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/spreadsheetId/edit#gid=0`.
+- **Sheet**: Choose a sheet in which you want to delete. 
+    - Select **From list** to choose the sheet title from the dropdown list, **By URL** to enter the url of the sheet, **By ID** to enter the `sheetId`, or **By Name** to enter the name of the sheet.
+    - You can find the `sheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/aBC-123_xYz/edit#gid=sheetId`. 
+- **Start Row Number** or **Start Column**: Enter the row number of column name to start deleting.
```
freakwriter, line 147:
```suggestion
- **Start Row Number** or **Start Column**: Enter the row number or column letter to start deleting.
```

docs/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/sheet-operations.md
```diff
@@ -0,0 +1,220 @@
...
+- **Credential to connect with**: Create or select an existing [Google Sheets credentials](/integrations/builtin/credentials/google/).
+- **Resource**: Select **Sheet Within Document**.
+- **Operation**: Select **Get Row(s)**.
+- **Document**: Choose a spreadsheet that contains the sheet you want to get rows from. 
+    - Select **From list** to choose the spreadsheet title from the dropdown list, **By URL** to enter the url of the spreadsheet, or **By ID** to enter the `spreadsheetId`. 
+    - You can find the `spreadsheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/spreadsheetId/edit#gid=0`. 
+- **Sheet**: Choose a sheet you want to read rows from.
+    - Select **From list** to choose the sheet title from the dropdown list, **By URL** to enter the url of the sheet, **By ID** to enter the `sheetId`, or **By Name** to enter the name of the sheet.
+    - You can find the `sheetId` in a Google Sheets URL: `https://docs.google.com/spreadsheets/d/aBC-123_xYz/edit#gid=sheetId`.
+- **Filters**: Set filters to return a limited set of results: By default, the node returns all rows in the sheet.
```
freakwriter, line 167:
```suggestion
- **Filters**: By default, the node returns all rows in the sheet. Set filters to return a limited set of results: 
```

docs/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/sheet-operations.md
```diff
@@ -0,0 +1,220 @@
...
+  - **General Formatting**: 
+    - **Values (unformatted)** (default): n8n removes currency signs and other special formatting. Data type remains as number.
+    - **Values (formatted)**: n8n displays the values as they appear in Google Sheets (for example, retaining commas or currency signs) by converting the data type from number to string.
+    - **Formulas**: n8n returns the formula. It doesn't calculate the formula output. For example, if a cell B2 has the formula `=A2`, n8n returns B2's value as `=A2` (in text). Refer to [About date & time values | Google Sheets](https://developers.google.com/sheets/api/guides/formats#about_date_time_values){:target=_blank .external-link} for more information. 
+  - **Date Formatting**: Refer to [DateTimeRenderOption | Google Sheets](https://developers.google.com/sheets/api/reference/rest/v4/DateTimeRenderOption){:target=_blank .external-link} for more information.   	- **Formatted Text** (default): As displayed in Google Sheets, which depends on the spreadsheet locale. For example `01/01/2024`.  	- **Serial Number**: Number of days since  December 30th 1899.
```
freakwriter, line 182:
```suggestion
  - **Date Formatting**: Refer to [DateTimeRenderOption | Google Sheets](https://developers.google.com/sheets/api/reference/rest/v4/DateTimeRenderOption){:target=_blank .external-link} for more information.
     	- **Formatted Text** (default): As displayed in Google Sheets, which depends on the spreadsheet locale. For example `01/01/2024`.
     	- **Serial Number**: Number of days since  December 30th 1899. 
```

freakwriter (changes requested):
@ayatnkw this is a massive amount of work! I like splitting it out based on operation type, and after seeing what's happening with the tables, I prefer the bullet point list for options for now--I suspect we'd spend a lot of time fussing with it to get it to display well, which isn't time well-spent right now.

I've added a number of suggested commits, let me know if you want to discuss any further!

nik8n:
Looks good to me. Two pieces of feedback:
1. I think we could still add a little more "FAQ" to the page.
2. On `Append an Array` I would mention the AI transform node and the split out node, no? Both of these options seem much easier than using a code node. And maybe we also want to get a little more into detail here and provide examples or a how-to? (kinda like the cookbook pages we sometimes have)

PR #2334, opened 2024-08-12
DOC-1015: Update Typeform credential to medium tier template

freakwriter:
Summary of changes:

- Converted bullet points to detailed numbered instructions
- Added required scopes for access token
- Removed prerequisites section


freakwriter:
I tested all the steps, so this should just require a formatting review. 😸

nik8n (commented):
I like that we don't have prerequesits here :P. I would probably always move the Related resources below the instructions on how to connect, but feel free to disagree.

ayatnkw (approved):

freakwriter:
@nik8n I've debated that change as we've moved along, but haven't felt strongly enough about it to make it yet. It's one of the things I'd like to review once we've gotten through the first round of credentials updates and we feel good about the rest of the changes.

PR #2336, opened 2024-08-12
DOC-1016: Update AWS credentials to medium tier template

freakwriter:
Removed separate subheadings to collapse into single list, added more detail on the steps involved.


freakwriter:
Should be a fairly fast review, as I just added more detail and streamlined the format.

ayatnkw (approved):

PR #2338, opened 2024-08-12
DOC-1018: Update HubSpot credentials to medium tier template

freakwriter:
Summary of changes:

- Convert high-level bullet point lists to more detailed numbered lists.
- Convert scopes lists to tables.
- Add link to migration guide for deprecated API key method.

### DIFF
--- a/docs/integrations/builtin/credentials/hubspot.md
+++ b/docs/integrations/builtin/credentials/hubspot.md
@@ -100,7 +100,7 @@ Some HubSpot accounts don't have access to all the scopes. HubSpot is migrating 
 
 --8<-- "_snippets/integrations/builtin/credentials/cloud-oauth-button.md"
 
-If you're self-hosting n8n, you'll need to configure OAuth2 from scratch by creating a new public app:
+If you're [self-hosting](/hosting/) n8n, you'll need to configure OAuth2 from scratch by creating a new public app:
 
 1. Log into your [HubSpot app developer account](https://developers.hubspot.com/){:target=_blank .external-link}.
 2. Select **Apps** from the main navigation bar.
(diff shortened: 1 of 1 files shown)


freakwriter:
@ayatnkw this one's probably a more involved review since the credentials are more complicated, but hopefully these changes make it cleaner/easier to use.

ayatnkw (approved):
looks good!

docs/integrations/builtin/credentials/hubspot.md
```diff
@@ -13,18 +13,14 @@ You can use these credentials to authenticate the following nodes:
...
 ## Supported authentication methods
 
-- App token: Used with [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node
-- Developer API key: Used with [HubSpot Trigger](/integrations/builtin/trigger-nodes/n8n-nodes-base.hubspottrigger/) node
-- OAuth2: Used with [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node
+- App token: Use with the [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node.
+- Developer API key: Use with the [HubSpot Trigger](/integrations/builtin/trigger-nodes/n8n-nodes-base.hubspottrigger/) node.
+- OAuth2: Use with the [HubSpot](/integrations/builtin/app-nodes/n8n-nodes-base.hubspot/) node.
 
 /// note | API key deprecated
```
ayatnkw, line 22:
```suggestion
/// warning | API key deprecated
```

freakwriter: @ayatnkw for deprecated things we usually just use a note--I used a warning for the X/Twitter thing since the behavior is wildly different between the two but everywhere else we use a note, which is why I went with it here. Though...in this case, since it's still in the UI, it probably does make sense to use warning since they won't get far if they use it. I'll commit your suggestion. :)

PR #2340, opened 2024-08-12
DOC-1010: Update X credentials to medium tier template

freakwriter:
Summary of changes:

- Replaced bullet point list with detailed numbered instructions on creating the Twitter app.
- Added note about when we released the V2 version of the X node so users can pick the correct node.


freakwriter:
@nik8n and @ayatnkw This review might be a little more involved since X/Twitter stuff is a bit odd since we still have the old auth method from older versions that folks may need.

Whether to document the method still is a grey area in our existing docs--when I asked Debs about when we officially sunset docs on things for old versions, it seemed to be a judgment call based on two factors:

- whether the thing was still discoverable in the UI
- how old the release was when it stopped being relevant

If it wasn't discoverable in the UI, and it applied to version 0.xxx of n8n, we generally removed the docs entirely. But if it was still in the UI and/or in version 1.xxx of n8n, we kept it.

This credential is a weird edge case. The OAuth credential ties to OAuth1.0a, which is only used with V1 of the X node. We released the updated version of that node in 0.236.0.

But you can still find that credential if you search from the Credentials interface (rather than from opening it within the X node itself). Jon and I had a brief convo about it [in Linear](https://linear.app/n8n/issue/NODE-1560/twitterx-credentials-remove-oauth-oauth1-option-from-credentials).

So for now I left it in but I moved it below the other auth method and tried to add some warnings about using it.

All of that said: I would really LIKE to remove this old auth method from the docs entirely, since it's not relevant for a lot of our current users. But because it's a bit of an edge case, that feels like a decision we should make collectively.

So: what do you think?

ayatnkw:
> @nik8n and @ayatnkw This review might be a little more involved since X/Twitter stuff is a bit odd since we still have the old auth method from older versions that folks may need.
> 
> Whether to document the method still is a grey area in our existing docs--when I asked Debs about when we officially sunset docs on things for old versions, it seemed to be a judgment call based on two factors:
> 
> * whether the thing was still discoverable in the UI
> * how old the release was when it stopped being relevant
> 
> If it wasn't discoverable in the UI, and it applied to version 0.xxx of n8n, we generally removed the docs entirely. But if it was still in the UI and/or in version 1.xxx of n8n, we kept it.
> 
> This credential is a weird edge case. The OAuth credential ties to OAuth1.0a, which is only used with V1 of the X node. We released the updated version of that node in 0.236.0.
> 
> But you can still find that credential if you search from the Credentials interface (rather than from opening it within the X node itself). Jon and I had a brief convo about it [in Linear](https://linear.app/n8n/issue/NODE-1560/twitterx-credentials-remove-oauth-oauth1-option-from-credentials).
> 
> So for now I left it in but I moved it below the other auth method and tried to add some warnings about using it.
> 
> All of that said: I would really LIKE to remove this old auth method from the docs entirely, since it's not relevant for a lot of our current users. But because it's a bit of an edge case, that feels like a decision we should make collectively.
> 
> So: what do you think?

If we're deprecating the OAuth eventually, I think we can remove the old auth method from the current docs. I think having more content on the page just makes it more confusing for new users. (But if we do keep it, I might change the warning note to `Deprecation warning` instead of `Recommended usage` ...) 

One question — in the UI we refer to the node as `X (formerly Twitter)`, maybe we should include the `(formerly Twitter)` part to make it more searchable? (Noticed other platforms and softwares also do this)

ayatnkw (approved):
Looks good, see comments re keep old OAuth or not.

freakwriter:
@ayatnkw thanks, that was all good feedback and I've incorporated all of it in some fashion.

PR #2341, opened 2024-08-13
Node overhaul: Schedule Trigger

ayatnkw:
### Changes
- Added custom cron expressions


docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger.md
```diff
@@ -22,13 +22,18 @@ Add **Trigger Rules** to determine when the trigger should run.
...
 Use the **Trigger Interval** to select the time interval unit of measure to schedule the trigger for. All other rule parameters depend on the interval you select. Choose from:
 
-* [**Seconds**](#seconds-trigger-interval)
-* [**Minutes**](#minutes-trigger-interval)
-* [**Hours**](#hours-trigger-interval)
-* [**Days**](#days-trigger-interval)
-* [**Weeks**](#weeks-trigger-interval)
-* [**Months**](#months-trigger-interval)
-* [**Custom (Cron)**](#custom-cron-interval)
+- [Schedule trigger](#schedule-trigger)
```
freakwriter, line 25: This just links people to the top of the page...is it supposed to be doing something else? I'm not sure it's needed.

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger.md
```diff
@@ -22,13 +22,18 @@ Add **Trigger Rules** to determine when the trigger should run.
...
+  - [Node parameters](#node-parameters)
```
freakwriter, line 26: This just links to the section that we're already in, so I'm not sure it's needed.

freakwriter:
@ayatnkw the Cron examples seem good, but it feels like you tried to turn the Node parameters list into a table of contents when it's not a table of contents--it's just a list of the various parameter options...?

freakwriter (commented):
I don't think the list in Node parameters should function as a table of contents. I'd likely stick with the "Choose from..." list and keep just the time intervals as options, and maybe add a sentence after that list to "Refer to [Templates and examples](#templates-and-examples) for more examples." or something.

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger.md
ayatnkw: good catch, thank you!

freakwriter:
> > I don't think the list in Node parameters should function as a table of contents. I'd likely stick with the "Choose from..." list and keep just the time intervals as options, and maybe add a sentence after that list to "Refer to [Templates and examples](#templates-and-examples) for more examples." or something.
> 
> I agree, I've reverted the changes I made, sorry about that!

No worries! We're still figuring this out as we go, and you made plenty of other changes here. :)

freakwriter:
@ayatnkw I'm not seeing any commits from my last review--is it possible you didn't push your changes?

PR #2349, opened 2024-08-15
Remove Ask AI options

Joffcom:
Removed AI config options and redirected to env options overview page to prevent 404s
Removed Ask AI section from HTTP Request node
Added note to the `1.50.0` release note so users know


freakwriter (approved):
Looks good, @Joffcom !

PR #2386, opened 2024-08-20
Add --no-template flag to reduce build time

ayatnkw:
All integration pages have a Workflow templates section with links to the most viewed templates in the templates library. But fetching the URLs for all integration pages takes a long time. There's probably a better implementation but wanted to bring it up. 

### Description
Add a `--no-template` flag to the mkdocs serve command which prevents API requests from fetching workflow templates every time it builds, replacing them with a placeholder URL (https://n8n.io/workflows/).

### Alternative solutions
- Caching workflow template URLs?

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,11 +2,19 @@ import requests
 import json
 import re
 import urllib.parse
+import sys
 
 def define_env(env):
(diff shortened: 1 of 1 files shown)


Joffcom (approved):
Love this, nice work.

PR #2387, opened 2024-08-20
update templates metadata

ayatnkw:
Related [Notion page](https://www.notion.so/n8n/SEO-for-docs-63cb35d9fb5647668c383848c92f7ee0?pvs=4)


freakwriter (approved):
Thanks for thinking of this and actually updating the templates. :)

PR #2396, opened 2024-08-22
Updated the wording around installing non-npm nodes in docker

dkarzon:
The docs weren't very clear about what files needed to be put into the docker container so it took me some time to figure it out.


freakwriter (approved):
@dkarzon thank you for submitting this PR to save others from the same confusion. 🥲

PR #2405, opened 2024-08-23
Update OpenWeatherMap credentials to medium tier template

freakwriter:
This setup is so simple it didn't seem worth numbered instructions, but I incorporated the prerequisites into the auth section and reformatted it mildly for readability.


nik8n:
We don't we do the usual numbered list approach here? Other services like Mistral also seem pretty easy. Or can users not create keys here?

freakwriter:
@nik8n calling me out on being a little lazy with my instructions... 😂 

Updated to include numbered instructions both referencing the original welcome email and how to create a new key from scratch.

nik8n (approved):

PR #2422, opened 2024-08-27
changes Function node to Code node

riascho:
we do not use 'Function Nodes' anymore


freakwriter (approved):
@riascho thanks for catching and updating this. I made one further tweak, to change it to a hyperlink to the Code node itself (because...why not?).

PR #2439, opened 2024-08-29
Improve instructions for Docker Compose users

gurbaxani:
Added steps to switch to the Docker user and navigate to the n8n-caddy folder.

### DIFF
--- a/docs/hosting/installation/docker.md
+++ b/docs/hosting/installation/docker.md
@@ -103,6 +103,12 @@ From your Docker Desktop, navigate to the **Images** tab and select **Pull** fro
 You can also use the command line to pull the latest, or a specific version:
 
 ```sh
+# Switch to the docker user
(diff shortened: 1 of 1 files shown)


Joffcom:
Hey @gurbaxani,

Sorry for the delay on this one, The steps you have put are specific to Docker compose using our install guide but placed in the Docker section so they sadly won't work.

We could include the folder change into the install guides for Hetzner and Digital Ocean which use the `n8n-caddy` folder, Swapping to the `docker` user is also optional and not a requirement so maybe that part can have a note to mention that you may need to change to the docker user if your user doesn't have permission to run the docker commands.

StarfallProjects:
Hiya! Thank you for the contribution!

As there's not been any activity for a while, I'm closing this. But feel free to reopen if you want to continue working on it.

PR #2447, opened 2024-09-03
Updates the intructions on how to get a chat ID for telegram

nik8n:
After seeing some errors from users trying to use the Telegram node, I decided to quickly update the help docs. I'm not sure about the format or even the branch name. Please help me @freakwriter :D


freakwriter (approved):
@nik8n I discovered there's a third way to get the ID, so I reworked your existing stuff slightly and added the third way, too. :)

nik8n (commented):
Thanks so much for rewriting. Looks much better now :)

PR #2451, opened 2024-09-04
doc: Add new parameter description

burivuhster:
The new option was added for Text Classifier Node in https://github.com/n8n-io/n8n/pull/10667
It is the same option as in the Sentiment Analysis Node.


freakwriter (approved):
@burivuhster thanks for adding this. Approved!

PR #2452, opened 2024-09-04
docs: Postgres connection timeout

ivov:
Ref https://github.com/n8n-io/n8n/pull/10670


ayatnkw (approved):

ivov:
Edited to reflect updated default: https://github.com/n8n-io/n8n/pull/10705

PR #2459, opened 2024-09-05
DOC-1047: Add Managing the Session ID section to Window Buffer Memory

freakwriter:
Added the section (with small tweaks to make our linting tool happier) as outlined [here](https://linear.app/n8n/issue/DOC-1047/add-more-context-about-the-sessionid-in-the-window-buffer-memory-tool).


freakwriter:
@nik8n you'd put this one into Triage, can you review?

nik8n:
@freakwriter the content looks good. Could we also put this into a separate page, like we did with telegram? Also, the part with `completions` etc, currently feels like it belongs to the managing the session ID section. Is that wanted?

freakwriter:
@nik8n Ah the AI glossary--I'd added a title to that snippet but this branch hadn't been updated, so it should be more clear what that section is now.

I've split the common issues out to a separate page. Let me know if you want any other changes here.

nik8n (approved):

PR #2471, opened 2024-09-12
DOC-1107: Update OpenAI Chat Model to high tier template

freakwriter:
Mostly just formatting changes here. The Chat Model didn't have errors in the common errors list.


freakwriter:
@nik8n let me know if there are other changes you'd like here.

PR #2472, opened 2024-09-12
DOC-1109: Update Gmail Trigger to high tier template

freakwriter:
Summary of changes:

- Moved Poll Times options into a snippet, since other triggers use the same setting.
- Added all parameters and options.
- Updated RSS feed read trigger to use same snippet.
- Added auth error.


nik8n:
Let's also have a separate common issues page here. Felt much cleaner. It also feels like there's something wrong with the formating. The poll options now are under common issues

nik8n:
The operations and options also still feel a little bit "much", no? Feel free to disagree

freakwriter:
@nik8n "a bit much"...meaning you preferred the bullet point list instead of the separate headings? I can shift back to that.

nik8n:
Sorry for the fluffy comms :D. With "a bit much" I mean the amount of content on the page. It felt a little chaotic to me with node params, node filters and all the different poll mode options being relatively close in section weights. I was thinking if we could/should hide some of this behind dedicated pages

freakwriter:
Ahh okay, that makes more sense. The Poll section is the one that makes the page feel gigantic, I think. Let me try moving that to its own page first and see what that looks like.

freakwriter:
@nik8n with the poll mode options on a new page, this one feels a lot more manageable (I also reverted the separate headings per parameter/option in case that was causing some of the visual overload). Let me know what you think.

nik8n (approved):
Yes, much better. Thank you!

PR #2484, opened 2024-09-17
DOC-1114: Update Manual Trigger node to critical tier template

freakwriter:
Summary of changes:

- Updated front matter and nav to current SEO-friendly templates.
- Reworked existing content for clarity.
- Added one common error people might run into.


docs/integrations/builtin/core-nodes/n8n-nodes-base.manualworkflowtrigger.md
```diff
@@ -1,14 +1,32 @@
 ---
```
imchairmanm, line 1: Are these title and description formats something we want to update in the trigger template?

imchairmanm (approved):

docs/integrations/builtin/core-nodes/n8n-nodes-base.manualworkflowtrigger.md
freakwriter: Good call.

PR #2485, opened 2024-09-17
DOC-1116: Google Sheets node updates

freakwriter:
Summary of changes:

- Move troubleshooting stuff to its own "Common issues" page
- Test using a cards layout for the operations list, instead of the bullet point list


freakwriter:
@nik8n I know we've both not loved the bullet point list of operations in the nodes. While I was adding a separate common issues page for Google Sheets, I played around with using a cards layout for the operations list instead.

Curious what you and @imchairmanm think--you should be able to compare this to the current [Google Sheets node doc](https://docs.n8n.io/integrations/builtin/app-nodes/n8n-nodes-base.googlesheets/). I can flip back to the bullet point list if we don't like this, but figured I'd give you an option. :)

imchairmanm:
@freakwriter The cards are interesting.  I don't think I've gathered all of my thoughts yet, but my initial reaction is:

* I think the categories ("Document" and "Sheet Within Document") look better as subheadings than bullets
* The title + link name in the cards seem a bit repetitive to me
* I think space that cards occupy would be a better tradeoff if we could figure out a way to make the whole element act as a link instead of just the URL. Not sure how well that fits with how the card mkdoc's card component is designed though.
* Personally, I think my eyes have a bit easier time parsing items that are only arranged on one axis (vertical or horizontal), so the cards are a bit more difficult to scan for me.

All of that being said, I don't know much about the pain points you all have experienced, so feel free to disregard.

freakwriter:
Okay good. I wasn't loving the cards either but wanted to present all the options, so I'll shift us back to the bullet point list.

freakwriter:
Okay, cards are removed and it sounded like the other changes were okay--would one of you be able to approve?

nik8n (approved):

PR #2490, opened 2024-09-19
Updated Navigation in Salesforce

curiousmarkus:
Since Winter'24 release, Salesforce introduced "External Client App" that will replace the current "connected Apps" at some point.


HumanistSerif:
This change has been addressed in #4251. Thanks for the feedback.

PR #2496, opened 2024-09-19
DOC-1001: Update Facebook Graph credentials to medium tier template

freakwriter:
Summary of changes:

- Removed prerequisites section
- Added more detailed instructions on creating the app and generating the token, similar to the other FB credentials. This one doesn't have a specific product since you can use it with many, so made these generic. Future work might add specific instructions for different apps.

### DIFF
--- a/docs/integrations/builtin/credentials/facebookgraph.md
+++ b/docs/integrations/builtin/credentials/facebookgraph.md
@@ -58,7 +58,9 @@ To create a Meta app:
 1. In the left menu, select **Add Product**.
 6. The **Add products to your app** page appears. Select the products that make sense for your app and configure them.
 
-Refer to Meta's [Create an app](https://developers.facebook.com/docs/development/create-an-app) documentation for more information.
+Refer to Meta's [Create an app](https://developers.facebook.com/docs/development/create-an-app){:target=_blank .external-link} documentation for more information on creating an app, required fields like the Privacy Policy URL, and adding products.
+
+For more information on the app modes and switching to **Live** mode, refer to [App Modes](https://developers.facebook.com/docs/development/build-and-test/app-modes){:target=_blank .external-link} and [Publish | App Types](https://developers.facebook.com/docs/development/release#app-types){:target=_blank .external-link}.
 
 ### Generate an App Access Token
 
--- a/docs/integrations/builtin/credentials/facebookgraph.md
+++ b/docs/integrations/builtin/credentials/facebookgraph.md
@@ -16,12 +16,6 @@ You can use these credentials to authenticate the following nodes:
 If you want to create credentials for the [Facebook Trigger](/integrations/builtin/trigger-nodes/n8n-nodes-base.facebooktrigger/) node, follow the instructions mentioned in the [Facebook App credentials](/integrations/builtin/credentials/facebookapp/) documentation.
 ///
 
-## Prerequisites
-
-- Create a [Facebook](https://www.facebook.com/){:target=_blank .external-link} account.
-- Sign up for [Meta for Developers](https://developers.facebook.com/){:target=_blank .external-link} with that account.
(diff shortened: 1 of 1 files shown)


docs/integrations/builtin/credentials/facebookgraph.md
```diff
@@ -32,7 +26,54 @@ Refer to [Meta's Graph API documentation](https://developers.facebook.com/docs/g
 
 ## Using app access token
 
-To configure this credential, you'll need:
+o configure this credential, you'll need a [Meta for Developers](https://developers.facebook.com/){:target=_blank .external-link} account and:
```
imchairmanm, line 29:
```suggestion
To configure this credential, you'll need a [Meta for Developers](https://developers.facebook.com/){:target=_blank .external-link} account and:
```

docs/integrations/builtin/credentials/facebookgraph.md
```diff
@@ -32,7 +26,54 @@ Refer to [Meta's Graph API documentation](https://developers.facebook.com/docs/g
...
+3. Select the **Use case** that aligns with how you wish to use the Facebook Graph API. For example, for products in Meta's **Business** suite (like Messenger, Instagram, WhatsApp, Marketing API, App Events, Audience Network, Commerce API, Fundraisers, Jobs, Threat Exchange, and Webhooks), select **Other**, then select **Next**.
+4. Select **Business** and **Next**.
+5. Complete the essential information:
+    * Add an **App name**.
+    * Add an **App contact email**.
+    * Here again you can connect to a business portfolio or skip it.
+1. Select **Create app**.
+1. The **Add products to your app** page opens.
+1. Select **App settings > Basic** from the left menu.
+1. Enter a **Privacy Policy URL**. (Required to take the app "Live.")
```
imchairmanm, line 55: This is one place where I'd typically consider linking to a doc from Meta that confirms this information and explains the requirements more in-depth. I'm unsure about if / how we do this at n8n though. Curious to hear your thoughts.

docs/integrations/builtin/credentials/facebookgraph.md
```diff
@@ -32,7 +26,54 @@ Refer to [Meta's Graph API documentation](https://developers.facebook.com/docs/g
...
+Next, create an app access token to use with your n8n credential and the products you selected:
+
+1. In a separate tab or window, open the [Graph API explorer](https://developers.facebook.com/tools/explorer/){:target=_blank .external-link}.
+2. Select the **Meta App** you just created in the **Access Token** section.
+3. In **User or Page**, select **Get App Token**.
+4. Select **Generate Access Token**.
+5. The page prompts you to log in and grant access. Follow the on-screen prompts.
+
+    /// warning | App unavailable
+    You may receive a warning that the app isn't available. Once you take an app live, there may be a few minutes' delay before you can generate an access token.
```
imchairmanm, line 74: Ah, this is what you mentioned the other day! Yeah, that makes perfect sense as an exception. Great example.

imchairmanm (approved):

docs/integrations/builtin/credentials/facebookgraph.md
freakwriter: @imchairmanm yes, that would be a good practice, if Meta had docs that were particularly clear. I've certainly done that other places. :) 

My dilemma was this: their docs officially say that ["Business" apps don't need to go into "Live" mode](https://developers.facebook.com/docs/development/release#app-types):
>Business apps do not have app modes and rely exclusively on access levels.

But in my testing, our credentials won't connect if the app isn't in Live mode. So I was hesitant to link to official docs that might mildly contradict what we're instructing folks to do. 

More info on the modes is available [here](https://developers.facebook.com/docs/development/build-and-test/app-modes).

My inclination would be to add both of these links in the "Refer to xxx" section as additional resources rather than interrupting the stream of instructions here, given my own misgivings about their docs. Wdyt?

This one's your call at this point.

PR #2497, opened 2024-09-20
Update section "Using API access token"

marceloschmidt:
Changes information about the host and port fields. Host should not include http or https protocol.


freakwriter (approved):
Thanks for adding these details and clarity, @marceloschmidt! I've made a couple small tweaks for our current style guide. We really appreciate submissions like this that help our community avoid confusion or headaches--thank you!

PR #2514, opened 2024-09-27
Added 429 advice for HTTP Request node

nik8n:
I have added advice on how to handle 429 errors in our HTTP request node. It would probably be nice to link to our cookbook section that mentions this as well... but I wasn't quite sure where to find this.

Also, I'm not a professional tech writer, so feel free to rip things apart @imchairmanm :D


nik8n:
Ah yeah, that's the one I meant.

imchairmanm (approved):

PR #2515, opened 2024-09-27
DOC-1124: Add Webhook URL section to common issues

freakwriter:
Per Nik's request, adding a section to common issues about the different Webhook URLs.


freakwriter:
@nik8n you'd requested that we add something to the Webhook common issues page about Test vs. Production URL. Can you take a look at this?

imchairmanm (approved):

PR #2460, opened 2024-09-05
DOC-979: Remove docs for deprecated OpenAI text completion & Google Palm models

freakwriter:
Deleted these files:

- /integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmopenai/
- /integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmchatgooglepalm/
- /integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmgooglepalm/

Set up redirects for each to the current nodes JP provided in Linear.

Deleted any cross-references to those pages from other docs.

Removed them from the navigation.

PR #2518, opened 2024-09-30
DOC-1134: Make Google AI credential title match app, finish trigger > Trigger cleanup

freakwriter:
Credentials were previously titled "Google AI", which doesn't match anything in the UI. Updated to Google Gemini(PaLM) to match app UI.

Also noticed some lingering triggers that were still lower-cased and tidied those up a bit, as well.


freakwriter:
@imchairmanm this was a note I'd left myself months ago and never moved into Linear, and it seemed easier to just do it than leave it as a ticket for you. :)

freakwriter:
Also moving this so it's a standalone credential, per [this convo from David](https://linear.app/n8n/issue/DOC-1129/cred-docs-for-google-gemini-model)

freakwriter:
The three deleted files were deleted [here](https://github.com/n8n-io/n8n-docs/pull/2460).

They were then erroneously revived, so this is me just re-tidying them up.

freakwriter:
@nik8n adding you as a reviewer since Justin is out. This PR is a little messy but here's what it's doing:

- Rename "Google AI" credentials to match the label we use in the UI
- Bumping the page so it's a standalone since it's way less complicated than the other Google services, updating all navigation and links for that
- Re-deleting three files I previously deleted in another PR that somehow got re-added to the project (probably me, probably a bad merge conflict resolution on one of the days I released a ton of things)
- Fixing any lingering cases of trigger not being capitalized when it's in the name of the node itself, which was a style change we recently made

nik8n (approved):

PR #2531, opened 2024-10-09
Add error to common issues for OpenAI for exceeding your acount quota

imchairmanm:
This adds an error to the OpenAI node common issues for when the connected OpenAI account has insufficient quota to process the request.

### DIFF
--- a/docs/integrations/builtin/app-nodes/n8n-nodes-langchain.openai/common-issues.md
+++ b/docs/integrations/builtin/app-nodes/n8n-nodes-langchain.openai/common-issues.md
@@ -129,13 +129,13 @@ There are two ways to work around this issue:
 
 ## Insufficient quota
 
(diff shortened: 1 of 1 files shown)


nik8n (changes requested):
I think there's more to do here, no @imchairmanm?

I'd add these two things:
1. Check that your organization limit is not met yet
2. Check that your project limit is not met yet

imchairmanm:
Hi @nik8n just a few questions to make sure I understand. When I looked up the 'Insufficient quota' error that Jason was seeing, all of the comments seem to be talking about payment credit.  Do the organization and project limits also throw the same error or does that belong in a separate section?

nik8n:
To my knowledge that will throw the same error, not 100% sure though

nik8n (approved):

PR #2549, opened 2024-10-15
Updated invalid URL for starter repo

xjrcode:
The URL for starter repository does not exists and returns error 404 so this PR fixes the invalid URL


Joffcom:
Hey @xjrcode,

The URL is a special one, What it does is starts the process to make a copy of template on your own GitHub account. It looks like you get a 404 if you are not already logged in which is an assumption we have made if you have got that far.

Maybe a better solution would be to add a note to make sure the user is logged in.

StarfallProjects:
Hiya! Thanks for the contribution. As the CLA is unsigned and there's not been any action on the feedback, I'm closing this for now. Feel free to reopen it if you want to continue working on it.

PR #2562, opened 2024-10-21
Old link was depreciated on langchain docs. Have updated to the one…

jason8n:
…that now works


imchairmanm (approved):
Thanks for catching!

PR #2569, opened 2024-10-24
Adjust common errors details for OpenAI node

imchairmanm:


burivuhster (approved):
Thanks, @imchairmanm ! I think this guide would be really helpful.

PR #2580, opened 2024-10-29
Add common issues to execute command node

imchairmanm:


nik8n (changes requested):
One quick feedback: Can we add the "Common issues" behind a separate page? It helps our AI to navigate and also makes the page less intimidating when coming to it for the first time.

imchairmanm:
Yeah, for sure. I was thinking since it was relatively short it could live on the same page, but that's a good reason to move it :+1: Care to take another look?

nik8n (approved):

PR #2586, opened 2024-10-31
Add common issues to if node

imchairmanm:


nik8n (commented):
I guess same here? These don't sound like real if-node problems but more referencing data problems

imchairmanm:
Closing this for now since [common issues with expressions](https://docs.n8n.io/code/cookbook/expressions/common-issues/) is now available and we've refocused the project on non-core nodes.

PR #2588, opened 2024-11-01
Redirect security page to main site

imchairmanm:


Johannesishere (approved):
Since the security page is now an external link in the nav, it doesn't show up in the index file anymore. Is this a problem? I would think not.

PR #2599, opened 2024-11-06
Update docker-compose.md

lazee486:
added a oneliner docker and compose install suggestion for fresh OS installs.   please check is appropriately placed.


StarfallProjects:
Hi @lazee486 thank you for the contribution, and my apologies for the very slow response.

I'm going to close this. In the Docker docs, it mentions this [convenience script](https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script) approach, but warns that "The convenience script isn't recommended for production environments", and also advises users to still check the preparatory steps, as well as familiarise themselves with potential risks and limitations.

Many of our users will simply follow our guide without doing further reading, and may well be using this in production, so I don't want to encourage them down a route that Docker doesn't advise. This sounds more like a step for experienced users/people who are already working with Docker in depth, and who will know when to use it (and when not to)

PR #2607, opened 2024-11-08
Add common issues to chat trigger node

imchairmanm:


nik8n (commented):
TBH I did not fully get the 2nd point. But maybe that is because I did not have the problem yet.

imchairmanm:
@nik8n Yeah, to be honest it's a bit unclear to me too.

I was attempting to try and resolve the conflict between [this solution](https://community.n8n.io/t/n8n-chat-trigger-failing-to-fetch-previous-messages-loadprevioussession/49015/11) (don't connect them to the same memory buffer node), with [what we say here](https://deploy-preview-2607--n8n-docs.netlify.app/integrations/builtin/core-nodes/n8n-nodes-langchain.chattrigger/#load-previous-session) (we recommend that they *do* connect everything to the same memory buffer node). I'm probably missing something though.

@jeanpaul is this something you could give some insight into?

jeanpaul:
I wrote this on Notion too:
> [The issue that that solution was for] was mostly because they’ve referred to nodes in their memory settings that would (for the chat-trigger) not be available yet. The tricky thing about re-using nodes is that expressions are evaluated (for these sub-nodes) at the root-node where it’s used. So, the same expression can have different values depending on where it’s used.

PR #2619, opened 2024-11-14
Correct Value 1

BaderAlrowaiei:
Corrected from {{$json["orderStatus"]}} to {{ $json.orderStatus }}.


Joffcom (approved):
This looks like a good change, thanks

PR #2632, opened 2024-11-19
Fix example of creating multiple items from single item

ihortom:
This comes from the question in the community forum, https://community.n8n.io/t/how-to-use-the-code-node-to-create-multiple-items-from-a-single-item/62189.


imchairmanm (approved):
Thanks @ihortom!

PR #2673, opened 2024-12-03
slack credentials may fail when token rotation turned on

riascho:
This PR adds information on why slack credentials may fail if the `token rotation` feature was turned on in the slack app. 
I made a snippet so it can be used in two places: the `slack credentials` page and the `slack trigger node` page, as these are the most common places that users might look for help. 

Please feel free to rephrase my wordings :) Thanks for the review!


imchairmanm:
Thanks @riascho! I pushed some small changes to try to make vale be quiet, but looks good otherwise. Thanks again!

PR #2679, opened 2024-12-05
fix(Form Trigger node): remove depreacted option

ayatnkw:
Using 'Respond to Webhook' Node option isn't shown in v2.2 + above and this pr removes it from the docs to prevent [confusion](https://community.n8n.io/t/respond-to-webhook-node-in-form-is-missing/63075).


imchairmanm (approved):
Looks good, thanks @ayatnkw

PR #2692, opened 2024-12-09
Accept text edits

evilnick:
The Vale `accept.txt` file takes regex, so I rationalised it a bit. It isn't entirely pointless! As well as being slightly less laborious, it is also a teeny tiny bit faster to process.

+ bonus typo fix


Joffcom (approved):
Makes perfect sense to me

imchairmanm (approved):
Awesome! This will definitely clean up the file a bit, which I'm happy about. :+1:

PR #2720, opened 2024-12-18
minor changes to longer-quickstart

evilnick:
Just a few teaks and clarifications.

@imchairmanm I couldn't think of an elegant way around hard-coding the step number in line 104,  but feel free to change

### DIFF
--- a/docs/try-it-out/longer-introduction.md
+++ b/docs/try-it-out/longer-introduction.md
@@ -93,13 +93,15 @@ Add the If node:
 		If you didn't follow the step in the previous section to run the NASA node, you won't see any data to work with in this step.
 		///
 
-    2. Change the comparison operation to **Contains**.
+    2. Change the comparison operation to **String>Contains**.
     3. In **Value 2**, enter **X**. This is the highest classification of solar flare. In the next step, you will create two reports: one for X class solar flares, and one for all the smaller solar flares.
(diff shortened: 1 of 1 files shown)


docs/try-it-out/longer-introduction.md
```diff
@@ -93,14 +93,16 @@ Add the If node:
 		If you didn't follow the step in the previous section to run the NASA node, you won't see any data to work with in this step.
 		///
 
-    2. Change the comparison operation to **Contains**.
+    2. Change the comparison operation to **String>Contains**.
```
imchairmanm, line 96:
```suggestion
    2. Change the comparison operation to **String > Contains**.
```
Maybe? I think this is the format we've used elsewhere (but it's likely to be inconsistent).

imchairmanm:
Yeah, regarding the step numbers, there's definitely some issues with auto-numbering sometimes.

We have [this extension enabled](https://python-markdown.github.io/extensions/sane_lists/) (you can see it in the `mkdocs.yml` file [here](https://github.com/n8n-io/n8n-docs/blob/e6df91412a071801f135bc9ca5a77910c19b335a/mkdocs.yml#L148)), but mainly this just allows us to hardcode the numbering if it's not doing things correctly, which wasn't possible in the vanilla config as far as I remember.

imchairmanm (approved):

PR #2721, opened 2024-12-19
Example of Digital Ocean

alnutile:
Just wanted to add an image to help with Digital Ocean setup


StarfallProjects:
Hiya! Thanks for the contribution, and sorry for the slow response.

I'm closing this: the CLA is unsigned, we generally keep use of screenshots minimal (we don't add examples like this generally), and this screenshot is pulling from an external source rather than included in the repo.

alnutile:
signed the CLA btw, just did not notice that.

PR #2722, opened 2024-12-19
docs: Remove sharing requirement for subworkflows

ivov:
Unneeded since https://github.com/n8n-io/n8n/pull/10893


imchairmanm (approved):
Thanks @ivov!

PR #2734, opened 2024-12-21
Update change-ownership-or-username.md

andrematzke:
I think, this fact is important to mention in the documentation. So one stays not behind confused.


evilnick:
@andrematzke Many thanks for this clarification - apologies for the delay in getting to it.

evilnick (approved):

PR #2749, opened 2025-01-13
Add a note about local ollama connection troubleshooting.

violabs-josh:
Added a note about using 127.0.0.1 instead of localhost if you encounter issues with node connecting.


violabs-josh:
Hello @evilnick . I totally understand the ceremony. Newer to the process also, so I will remember that. All set!

evilnick (approved):
+1

PR #2759, opened 2025-01-16
Add info on recurring events to Google Calendar docs

imchairmanm:
Also took the opportunity to document some of the operations and options more since we previously did not document a lot of functionality.


evilnick (approved):
:cake:

PR #2760, opened 2025-01-16
Add cred only node for Microsoft Azure Monitor

Joffcom:
Adds docs for credential only node, Dev PR: https://github.com/n8n-io/n8n/pull/12645

API Docs: https://learn.microsoft.com/en-us/azure/azure-monitor/azure-monitor-rest-api-index
Credential Docs: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/api/access-api?tabs=rest#set-up-authentication

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -1096,6 +1096,7 @@ nav:
         - integrations/builtin/credentials/medium.md
         - integrations/builtin/credentials/messagebird.md
         - integrations/builtin/credentials/metabase.md
+        - integrations/builtin/credentials/microsoftazuremonitor.md
         - integrations/builtin/credentials/microsoft.md
         - integrations/builtin/credentials/microsoftentra.md
         - integrations/builtin/credentials/microsoftsql.md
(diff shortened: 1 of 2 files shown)


mkdocs.yml
```diff
@@ -1096,6 +1096,7 @@ nav:
         - integrations/builtin/credentials/medium.md
         - integrations/builtin/credentials/messagebird.md
         - integrations/builtin/credentials/metabase.md
+        - integrations/builtin/credentials/microsoftazuremonitor.md
```
imchairmanm, line 1099: We should probably adjust the order slightly so that this is after the generic Microsoft creds.

imchairmanm (approved):

mkdocs.yml
Joffcom: That is a good idea, will tweak shortly.

Joffcom: @imchairmanm done

mkdocs.yml
```diff
@@ -1097,6 +1097,7 @@ nav:
         - integrations/builtin/credentials/messagebird.md
         - integrations/builtin/credentials/metabase.md
         - integrations/builtin/credentials/microsoft.md
+	- integrations/builtin/credentials/microsoftazuremonitor.md
```
imchairmanm, line 1100:
```suggestion
        - integrations/builtin/credentials/microsoftazuremonitor.md
```

PR #2787, opened 2025-01-24
DOC-1329: Update AWS Secrets Manager docs

MarcL:
- Updates the AWS Secret Manager docs to clarify the IAM actions needed to access the secrets.
- Added an example of a more restrictive IAM policy example to show users how to avoid exposing all the secrets in their secret manager to n8n.

Fixes: DOC-1329
Related: PAY-2426

### DIFF
--- a/docs/external-secrets.md
+++ b/docs/external-secrets.md
@@ -26,15 +26,15 @@ Your secret names can't contain spaces, hyphens, or other special characters. n8
 1. Select **Set Up** for your store provider.
 1. Enter the credentials for your provider:
 	* Azure Key Vault: Provide your **vault name**, **tenant ID**, **client ID**, and **client secret**. Refer to the Azure documentation to [register a Microsoft Entra ID app and create a service principal](https://learn.microsoft.com/en-us/entra/identity-platform/howto-create-service-principal-portal){:target=_blank .external-link}. n8n supports only single-line values for secrets.
-	* AWS Secrets Manager: provide your **access key ID**, **secret access key**, and **region**. The IAM user must have the `secretsmanager:ListSecrets` and `secretsmanager:BatchGetSecretValue` permissions.
+	* AWS Secrets Manager: provide your **access key ID**, **secret access key**, and **region**. The IAM user must have the `secretsmanager:ListSecrets`, `secretsmanager:BatchGetSecretValue`, and `secretsmanager:GetSecretValue` permissions.
 
-		Example policy:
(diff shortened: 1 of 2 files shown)


docs/external-secrets.md
```diff
@@ -50,9 +50,11 @@ Your secret names can't contain spaces, hyphens, or other special characters. n8
 		}
 		```
 
-		If you'd like to be more restrictive and avoid n8n having access to all of your secrets, you'll still need to allow `secretsmanager:ListSecrets` and `secretsmanager:BatchGetSecretValue` access to all resources. This doesn't allow access to the secret values but is needed to retrieve any ARN-scoped secrets. You will need to scope `secretsmanager:GetSecretValue` to the specific Amazon Resource Names (ARNs) for the secrets you wish to share with n8n. Ensure you use the correct region and account ID in each resource ARNs. You can find the ARN details in the AWS dashboard for your secrets.
+		You can also be more restrictive and give n8n access to select AWS Secret Manager secrets. You still need to allow the `secretsmanager:ListSecrets` and `secretsmanager:BatchGetSecretValue` permissions to access all resources. These permissions allow n8n to retrieve ARN-scoped secrets, but don't provide access to the secret values.
```
MarcL, line 53: > You can also be more restrictive and give n8n access to select AWS Secret Manager secrets.

We're trying to say that we could allow access to specific secrets and not all here and I don't think that's clear in that sentence. Perhaps this?

> You can also be more restrictive and give n8n access to select selected AWS Secret Manager secrets.

MarcL (commented):
@imchairmanm - Minor clarification. What do you think?

docs/external-secrets.md
imchairmanm: Hmm, sorry @MarcL I don't know if I quite understand your comment. Is it that "select" in the original context doesn't necessarily communicate "a specific selection" (that was my intent) or is it something else that's confusing?

MarcL: I had a typo in my suggestion. 🤦‍♂️

I meant to say:

> You can also be more restrictive and give n8n access to select specific AWS Secret Manager secrets.

I didn't think it was as clear that you can scope it down to specific secrets that you expose to n8n and you don't have to expose them all with the general wildcard `*`.

Does that make sense?

imchairmanm: Yes! Perfectly. Thanks for the added context, I'll make this change and publish.

imchairmanm (approved):

PR #2791, opened 2025-01-28
Add keys extension

evilnick:
This extension renders keys with the appropriate keycap symbols such as <kbd>Tab</kbd>

Will need to replace these as we find them, fixed one entry as an example


evilnick:
@imchairmanm here's the example: https://deploy-preview-2791--n8n-docs.netlify.app/courses/level-one/chapter-1/#finding-nodes

imchairmanm (approved):
This looks good to me. Let's use it :+1:

evilnick:
I get special bonus points for adding it to the wiki too

PR #2800, opened 2025-01-31
Add clarity around variables used with scheduled triggers

MarcL:
Add some clarity around how variables don't work as expected with scheduled triggers:

- Update scheduled trigger page to include a warning around using variables in cron expressions
- Update scheduled triggers common issues page to include details of variables not working as expected

### DIFF
--- a/docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
+++ b/docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
@@ -38,3 +38,7 @@ To set the timezone for an individual workflow:
 1. Select **Settings**.
 1. Change the **Timezone** setting.
 1. Select **Save**.
+
+### Variables not working as expected
+
+While variables can be used in the scheduled trigger, their values only get evaluated when the workflow is activated. After activating the worfklow, you can alter a variable's value in the settings but it won't change how often the workflow runs. To work around this, you must stop and then re-activate the workflow to apply the updated variable value.
--- a/docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
+++ b/docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
@@ -42,3 +42,9 @@ To set the timezone for an individual workflow:
 ### Variables not working as expected
 
 While variables can be used in the scheduled trigger, their values only get evaluated when the workflow is activated. After activating the worfklow, you can alter a variable's value in the settings but it won't change how often the workflow runs. To work around this, you must stop and then re-activate the workflow to apply the updated variable value.
+
+### Changing the trigger interval
+
+You can update the scheduled trigger interval at any time but it only gets updated when the workflow is activated. If you change the trigger interval after the workflow is active, the changes won't take effect until you stop and then re-activate the workflow.
+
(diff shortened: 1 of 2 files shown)


evilnick:
@MarcL Many thanks for your efforts here - I'm going to add some comments/suggestions as I see there have been some questions and issues related to this and it would be good to clear them all up at once.

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
```diff
@@ -38,3 +38,7 @@ To set the timezone for an individual workflow:
 1. Select **Settings**.
 1. Change the **Timezone** setting.
 1. Select **Save**.
+
+### Variables not working as expected
+
+While variables can be used in the scheduled trigger, their values only get evaluated when the workflow is activated. After activating the worfklow, you can alter a variable's value in the settings but it won't change how often the workflow runs. To work around this, you must stop and then re-activate the workflow to apply the updated variable value.
```
evilnick, line 44: Some of the queries I have seen have been along the lines of "If i have a schedule of 1 hour and I edit the workflow  and save it, does the one hour reset to  from the time I saved it?" It would be good if we could answer that here too!

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/index.md
```diff
@@ -100,6 +100,10 @@ To generate a Cron expression, you can use [crontab guru](https://crontab.guru){
 |Quarterly|`0 0 1 1,4,7,10 *`|At midnight on the 1st of January, April, July, and October.|
 <!-- vale from-write-good.Weasel = YES -->
 
+/// warning | Using variables in the Cron expression
+While variables can be used in the scheduled trigger, their values only get evaluated when the workflow is activated. If you alter a variable's value in the settings after a workflow is activated, the changes won't alter the cron schedule.
```
evilnick, line 104: Maybe add something like:
```suggestion
While variables can be used in the scheduled trigger, their values only get evaluated when the workflow is activated. If you alter a variable's value in the settings after a workflow is activated, the changes won't alter the cron schedule. To re-evaluate the variable, set the workflow to **Inactive** and then back to **Active** again
```

evilnick (changes requested):
Thanks again @MarcL. Just a few comments to hopefully make this even more useful!

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/index.md
MarcL: Done.

docs/integrations/builtin/core-nodes/n8n-nodes-base.scheduletrigger/common-issues.md
evilnick: Yes, as a separate entry. Sorry, i didn't mean to extend the scope of this but it just reminded me when we were here

MarcL: No problem. I'll sort that for you.

evilnick: Thanks so much!

evilnick (approved):

PR #2807, opened 2025-02-03
Update nextcloud.md: replace "dav" with "webdav"

jtmusson:
Following response in this issue by Joffcom: https://github.com/n8n-io/n8n/issues/8802

This states that the Nextcloud node expects the URL to include `/webdav/`, where the documentation states `/dav/`. This PR submits a change to amend the documentation.


evilnick:
Thanks @jtmusson, this must've been overlooked before, so I'm grateful for the help in keeping the docs accurate!

evilnick (approved):
:+1:

PR #2813, opened 2025-02-04
add content/priority to frontmatter

evilnick:
Some docs were missing regular sections of frontmatter

### DIFF
--- a/docs/integrations/builtin/core-nodes/n8n-nodes-base.jwt.md
+++ b/docs/integrations/builtin/core-nodes/n8n-nodes-base.jwt.md
@@ -3,6 +3,8 @@
 title: JWT
 description: Documentation for the JWT node in n8n, a workflow automation platform. Includes guidance on usage, and links to examples.
 contentType: [integration, reference]
+contentType: [integration, reference]
+priority: medium
 ---
 
 # JWT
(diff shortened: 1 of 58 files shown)


docs/integrations/builtin/core-nodes/n8n-nodes-base.jwt.md
```diff
@@ -3,6 +3,8 @@
 title: JWT
 description: Documentation for the JWT node in n8n, a workflow automation platform. Includes guidance on usage, and links to examples.
 contentType: [integration, reference]
+contentType: [integration, reference]
```
imchairmanm, line 6: Looks like this is maybe duplicating the above.

docs/integrations/builtin/credentials/huggingface.md
```diff
@@ -3,6 +3,8 @@
 title: Hugging Face credentials
 description: Documentation for the Hugging Face credentials. Use these credentials to authenticate Hugging Face in n8n, a workflow automation platform.
 priority: medium
+contentType: [integration, reference]
+priority: medium
```
imchairmanm, line 7: Looks like this is already above as well.

evilnick: :+1:

PR #2816, opened 2025-02-05
add pageinfo.py tool

evilnick:
Just a useful script for spitting out wordcounts and content

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,8 @@ __pycache__/
 
 ## Ignore pyenv configuration
 .python-version
+
+## Ignore ephemeral doc-tool output
+
+output.csv
+_doctools/*.csv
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -16,3 +16,8 @@ __pycache__/
 
 ## Ignore pyenv configuration
 .python-version
+
+## Ignore ephemeral doc-tool output
+
+output.csv
```
imchairmanm, line 22:
```suggestion
pageinfo.csv
```

Maybe? Just trying to match the default from the script, unless this is for something else.

imchairmanm (approved):
Super cool! This will be really nice to have. Thanks for making it!

.gitignore
evilnick: it **is** for something else, but i haven't finished making it yet. I'll leave it in for now :)

imchairmanm: Oooooh, mysterious! Can't wait to see :smile:

PR #2821, opened 2025-02-06
Bump Python version to 3.12

imchairmanm:
Bumping the version of Python to something more modern. [Python 3.8 reached end-of-life in October, 2024](https://devguide.python.org/versions/), so it's not a good idea to continue using it.

I use a Python 3.12 version for the site locally, so this PR is partially to make sure that Netlify can build and deploy it without issue.


imchairmanm:
@evilnick This is a follow up to the Material for MkDocs update. As discussed, we might want to defer merging until we have had time to verify that the previous update is fit for purpose.

evilnick (approved):

PR #2822, opened 2025-02-06
Upgrade mkdocs-macros-plugin

imchairmanm:
This bumps the [`mkdocs-macros-plugin` version](https://github.com/fralau/mkdocs-macros-plugin/releases). The new version doesn't necessarily seem to focus on performance improvements, but is a bit more well-tested.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 mkdocs-exclude==1.0.2
 mkdocs-glightbox
-mkdocs-macros-plugin==1.0.4
+mkdocs-macros-plugin==1.3.7


imchairmanm:
@evilnick again, this is just a follow up to the Material for MkDocs PR. We'll probably want to defer this so we can gain more confidence in the previous upgrade first.

evilnick:
@imchairmanm agreed. let's park this for now, but great work in getting all this seen to. 
I'm wondering whether we could do with a permanent staging branch, like 'Next'? Would be useful to be able to pre-fill with docs updates for stuff too rather than have everything lag behind

StarfallProjects:
Closing in favour of https://github.com/n8n-io/n8n-docs/pull/3444

PR #2824, opened 2025-02-06
Fix broken links identified by new build output

imchairmanm:
A nice enhancement from the MkDocs version upgrade is that the internal link checker is now spitting out broken internal links. These are the main ones.

The one I couldn't fix was this:

```
INFO    -  Doc file 'privacy-security/privacy.md' contains a link '#__consent', but there is
           no such anchor on this page.
```

This seems to be a special link to get the consent box to pop up again, giving users the chance to change their settings. I need to look into whether we can somehow mark this so the link checker ignores it.


evilnick:
Nice!

evilnick (approved):
:cake:

PR #2835, opened 2025-02-11
Remove orbit page

evilnick:
The page for the orbit node is redundant as the service closed long ago. 
There is already a redirect in place, so no point in keeping it around.


imchairmanm (approved):
Good idea!

PR #2844, opened 2025-02-13
Doc 1354 fromai2

evilnick:
- adds a snippet indicating page is for a node which can be used as an AI tool and pointing to the $fromAI docs
- adds this snippet to relevant node docs


imchairmanm (approved):
Looks good! :+1:

PR #2845, opened 2025-02-14
Adds additional info about the form description field

imchairmanm:
This also closes DOC-1334


imchairmanm:
@evilnick On second thought, this one is small enough that I'm actually just going to push it. We can update later if necessary of course.

PR #2847, opened 2025-02-14
Improve first two pages identified by feedback widget

imchairmanm:
Includes changes to the following pages:

* Airtable Trigger node: builds out some of the information more in line with our current templates.
* Loop Over Items node: explains some of the caveats and options just a bit more clearly. Replace an image with an embedded workflow.

It looks like Johannes' recent commit also reverted the update to the insiders submodule, so this one also resets that to the commit we recently updated to.

### DIFF
--- a/docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
+++ b/docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
@@ -12,7 +12,7 @@ The Loop Over Items node helps you loop through data.
 
 The node saves the original incoming data, and with each iteration, returns a predefined amount of data through the **loop** output.
 
-When the node execution completes, it combines all the data and returns it through the **done** output.
+When the node execution completes, it combines all of the processed data and returns it through the **done** output.
 
 ## When to use the Loop Over Items node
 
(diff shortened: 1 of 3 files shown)


docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
```diff
@@ -14,6 +14,15 @@ The node saves the original incoming data, and with each iteration, returns a pr
 
 When the node execution completes, it combines all the data and returns it through the **done** output.
```
evilnick, line 15:
```suggestion
When the node execution completes, it combines all of the processed data and returns it through the **done** output.
```
?

docs/integrations/builtin/core-nodes/n8n-nodes-base.splitinbatches.md
```diff
@@ -24,11 +33,9 @@ Enter the number of items to return with each call.
 
 ### Reset
 
-If turned on, the node will reset with the current input-data newly initialized with each loop.
+If turned on, the node will reset with the current input-data newly initialized with each loop. Use this when you want the Loop Over Items node to treat incoming data as a new set of data instead of a continuation of previous items.
```
evilnick, line 36: could we give a hypothetical example of this?

PR #2860, opened 2025-02-19
Update templates to link to markdown files directly instead of relative URL links

imchairmanm:
This is a follow up task after changing our repo from relative URL links (like `[link](/path/to/page)`) to absolute file links from the docs directory (like `[link](/path/to/file.md)`).


evilnick (approved):
Thanks!

PR #2873, opened 2025-02-20
Add scheduled external linkcheck

evilnick:
- uses lychee action to run a scheduled linkcheck for external links
- excludes internal links to docs.n8n.io
- excludes file:// and mail links

### DIFF
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -2,9 +2,11 @@ name: Lychee external linkcheck
 on:
   push:
     branches:
-      - main  # Adjust as needed
+      - main  
   pull_request:
-
+  schedule:
+    - cron: "18 7 * * *"  
+  workflow_dispatch: 
 jobs:
   link-check:
     runs-on: ubuntu-latest
--- a/.github/workflows/lychee.yml
+++ b/.github/workflows/lychee.yml
@@ -1,0 +1,19 @@
+- name: Restore lychee cache
+  id: restore-cache
+  uses: actions/cache/restore@v4
+  with:
+    path: .lycheecache
+    key: cache-lychee-${{ github.sha }}
+    restore-keys: cache-lychee-
+
+- name: Run lychee
+  uses: lycheeverse/lychee-action@v2
+  with:
+    args: "--base . --cache --max-cache-age 1d ."
+
+- name: Save lychee cache
+  uses: actions/cache/save@v4
+  if: always()
+  with:
+    path: .lycheecache
+    key: ${{ steps.restore-cache.outputs.cache-primary-key }}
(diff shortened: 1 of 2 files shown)


.github/workflows/lychee.yml
```diff
@@ -0,0 +1,24 @@
...
+  link-check:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Run Lychee Link Checker
+        uses: lycheeverse/lychee-action@v1
+        with:
+          args: --config lychee.toml ./docs
```
imchairmanm, line 20: Should we think about adding configuration for the `_snippets` directory too?

lychee.toml
```diff
@@ -0,0 +1,126 @@
...
+# accept = "200, 429"
+accept = ["200", "429"]
+
+# Proceed for server connections considered insecure (invalid TLS).
+insecure = false
+
+# Only test links with the given schemes (e.g. https).
+# Omit to check links with any other scheme.
+# At the moment, we support http, https, file, and mailto.
+scheme = ["https", "http"]
```
imchairmanm, line 63: Probably doesn't make sense to do as part of this, but we should also probably take a look and change any HTTP links to HTTPs in the repo. I did a quick look and there are a number (most I think automatically redirect to HTTPS, but I still think it'd be good to update).

lychee.toml
```diff
@@ -0,0 +1,126 @@
...
+
+# Ignore case of paths when matching glob patterns.
+glob_ignore_case = false
+
+# Exclude URLs and mail addresses from checking (supports regex).
+exclude = [
+  '^https://www\.linkedin\.com', 
+  '^https://web\.archive\.org/web/',
+  '^https://www.notion.so/n8n/Frontmatter-432c2b8dff1f43d4b1c8d20075510fe4',
+  '^https://docs.n8n.io/*'
```
imchairmanm, line 102: How does this last one work exactly?

imchairmanm (approved):
From what I can see, this looks reasonable. I think you might have to walk me through how to use this, but approving now to not block.

lychee.toml
evilnick: Aha! I set it up so that the internal links get rewritten as full links with the site repo, then I set this to ignore them:
 - it would use extra bandwidth we don't need currently
 - we will check the internal links with mkdocs anyway
 - finally, none of them work because they all have '.md' at the end.

lychee.toml
evilnick: agreed. we can probably do a local manual run of lychee to detect those

.github/workflows/lychee.yml
evilnick: Yes. i has trouble getting that to parse satisfactorily initially, so I left it off for now. I think there are some settings to tweak that will make it work

evilnick:
@imchairmanm There are going to need to be tweaks for sure. Turning down the aggressiveness might help with some of the github rate limiting i saw on some manual runs. There are probably other things we definitely want to exclude.
I can make a variation of this to check PRs also

PR #2879, opened 2025-02-21
Update license proxy to mention it should be lowercase

Joffcom:
The `http_proxy_license_server` and `https_proxy_license_server` need to be lowercase `http_proxy` and `https_proxy`


imchairmanm (approved):
Looks fine to me. Thanks Jon!

PR #2903, opened 2025-02-25
Add more specifics about manual, partial, and production executions

imchairmanm:


docs/workflows/executions/manual-partial-and-production-executions.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Manual executions
+
+Manual executions allow you to run workflows directly from the [canvas](/glossary.md#canvas-n8n) to test your workflow logic. These executions are "ad-hoc" — they run only when you manually select the **Test workflow** button.
+
+Manual executions make building workflows easier by allowing you to iteratively test as you go, following the flow logic and seeing data transformations. You can test conditional branching, data formatting changes, and loop behavior by providing different input items and modifying node options.
+
+/// note | Pinning execution data
+When performing manual executions, you can use [data pinning](/data/data-pinning.md) to "pin" or "freeze" the output data of a node. On future runs, instead of executing the pinned node, n8n will substitute the pinned data and continue following the flow logic. This allows you to iterate without operating on variable data or repeating queries to external services. Production executions ignore all pinned data.
```
evilnick, line 19: I don't want to go overboard on this note, but it might be worth briefly mentioning editing pinned data (https://docs.n8n.io/data/data-editing/) too.

docs/workflows/executions/manual-partial-and-production-executions.md
```diff
@@ -0,0 +1,54 @@
...
+///
+
+## Partial executions
+
+Clicking the **Test workflow** button at the bottom of the workflow in the **Editor** tab manually runs the entire workflow. You can also perform partial executions to run specific steps in your workflow. Partial executions are manual executions that only run a subset of your workflow nodes.
+
+To perform a partial execution, select a node, open its detail view, and select **Test step**. This executes the specific node and, if required, any preceding nodes required to fill in its input data. You can also temporarily disable specific nodes in the workflow chain to avoid interacting with those services while building.
```
evilnick, line 26: too many 'required's - maybe just omit the first?
```suggestion
To perform a partial execution, select a node, open its detail view, and select **Test step**. This executes the specific node and any preceding nodes required to fill in its input data. You can also temporarily disable specific nodes in the workflow chain to avoid interacting with those services while building.
```

evilnick (approved):
just some suggestions

docs/workflows/executions/manual-partial-and-production-executions.md
imchairmanm: Oooh, good catch :+1:

docs/workflows/executions/manual-partial-and-production-executions.md
imchairmanm: No, I think that's a great call. Adding that and a break to try to maintain readability.

```suggestion
When performing manual executions, you can use [data pinning](/data/data-pinning.md) to "pin" or "freeze" the output data of a node. You can optionally [edit the pinned data](https://docs.n8n.io/data/data-editing/) as well.

On future runs, instead of executing the pinned node, n8n will substitute the pinned data and continue following the flow logic. This allows you to iterate without operating on variable data or repeating queries to external services. Production executions ignore all pinned data.
```

PR #2905, opened 2025-02-26
Add example with a prefix in the expression for $fromAI.

jeanpaul:


docs/advanced-ai/examples/using-the-fromai-function.md
```diff
@@ -86,6 +86,12 @@ To dynamically populate the number of items you have in stock, you could use a `
 $fromAI("numItemsInStock", "Number of items in stock", "number", 5)
 ```
 
+If you only want to fill in parts of the parameter with a dynamic value from the model, you can use it as a normal expression as well. For example, when you are trying to fill in the subject of an e-mail, but always want to pre-fix that with 'Generated by AI:', you could use the following expression in the `subject` parameter:
```
imchairmanm, line 89:
```suggestion
If you only want to fill in parts of a field with a dynamic value from the model, you can use it in a normal expression as well. For example, if you want the model to fill out the `subject` parameter for an e-mail, but always pre-fix the generated value with the string 'Generated by AI:', you could use the following expression:
```

imchairmanm (approved):

PR #2915, opened 2025-02-27
Update api reference and styling

netroy:


MarcL:
@netroy ❤️  Thanks for doing this! Looks much better from just the minor change of the font.

CleanShot 2025-02-27 at 12 46 24

docs/api/api-reference.md
```diff
@@ -9,5 +9,22 @@ contentType: reference
 ---
 
 
-<redoc theme='{"typography": {"font-family":"Open sans", "headings":{"fontFamily":"Open sans"}}}' spec-url="/api/v1/openapi.yml"></redoc>
-<script src="https://cdn.jsdelivr.net/npm/redoc@latest/bundles/redoc.standalone.js"> </script>
+<redoc
+  spec-url="/api/v1/openapi.yml"
+  disable-search
```
imchairmanm, line 14: No strong feelings really, but curious why we decided to disable this.

imchairmanm (approved):

docs/api/api-reference.md
netroy: I've tried using it in the past, and the UX isn't great. 
I prefer no search to bad search, especially when the list isn't that long, and people can just use their browser search.

imchairmanm: Ah, gotcha. I think that's fair. Thanks for the explanation :+1:

PR #2917, opened 2025-02-27
Add makefile

Joffcom:
Adds a `Makefile` to check Python version, set up venv and install requirements, Also updates contributing guide to cover the new process.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ venv
 env
 site
 __pycache__/
+n8ndocs
 
 ## Ignore pyenv configuration
 .python-version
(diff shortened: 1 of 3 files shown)


evilnick:
Looks promising - will have to test it out a few places

StarfallProjects:
@Joffcom couple of questions:

1. Would need people to install Make (it's not there by default on Windows at least)
2. Potentially introduces confusion, if we're encouraging users to use `make run`, but in practice most of the time they'll need varied commands for testing depending on how much/how slow they want to build.

What's the benefit of using the Make file and adding another tool folks need to install? Happy to test it and use it if it's worth it.

Joffcom:
I think back when I added it there was some confusion on getting up and running, this gives us a consistent approach to it but would require make to be available.

There may be a better way to handle this

StarfallProjects:
@Joffcom I'm torn because I'm aware using makefiles on Python projects is a pretty standard good practice. Equally, it's one more thing to install, and I'm wondering if we can just make the README / internal docs clearer? Either way we have to handle supporting OS differences (I don't think the current commands will activate venv on Windows, for instance)

PR #2923, opened 2025-02-28
fix PostgreSQL example

evilnick:
simple one-line fix that gets around some user-reported issues with the example


imchairmanm (approved):
Thanks for taking a look at this :+1:

PR #2929, opened 2025-03-03
Update n8n-nodes-base.googleanalytics.md - possible typo

bangank36:
## Summary
- There seems to be a typo on line before `Operations`, this PR was supposed to remove it

## Reference
N/A


imchairmanm (approved):
@bangank36 I think you're right. Good catch and thanks for the PR! :+1:

bangank36:
> @bangank36 I think you're right. Good catch and thanks for the PR! 👍

Superb, thanks for approving

PR #2937, opened 2025-03-04
Updates index.md to fix verb conjugation

jpgninja:
Small change to fix typo. Original:

> This makes it useful for build a workflow to process data and return the results, like an API endpoint.

Update:

> This makes it useful for building a workflow to process data and return the results, like an API endpoint.


imchairmanm (approved):
Great catch, thanks for the PR!

PR #2940, opened 2025-03-06
Update chapter-5.3 - Fix node name

Linek:
I believe this should be an "IF node" there instead of the HTTP Node


Mohamed3nan:
I was going to commit that, yes indeed I think it should be "If"

imchairmanm (approved):
Thanks for the PR!

PR #2950, opened 2025-03-10
Add notes on moving credentials with workflow

imchairmanm:


jcn8n (commented):
@imchairmanm just had a look at these changes. Couple of small things. One that the wording "executes the same in its new context" wasn't that clear to me. I wonder if we need to say something more explicit like "ensuring a workflow continues to have access to the credential it needs to execute" (Maybe wording not quite right though for that)

Secondly, "When moving a workflow with credentials, you can choose to share credentials that you own" strictly speaking it isn't just whether you 'own' credentials but if you have the permission to share them on. So a instance owner/admin could choose to share any credentials on an instance even if they don't own them. So we should probably refine this to say something like that "you own or have the permission to share."

imchairmanm:
@jcn8n Thanks for the review! I can definitely work on those points. I appreciate you taking a look :+1:

evilnick (approved):
one tiny note

imchairmanm:
Okay, going to merge this one since I haven't had any further feedback.

PR #2953, opened 2025-03-11
Update running workers in queue mode

Joffcom:
Makes it clearer that executions_mode needs to be set for workers.


imchairmanm (approved):
Looks good, thanks Jon!

PR #2961, opened 2025-03-13
Update API note

evilnick:
Just a quick repeat of the API note and a resequence of the instructions.


evilnick:
:cowboy_hat_face:

PR #2982, opened 2025-03-17
Update lychee.yml

evilnick:
Target to specific SHA


evilnick:
@imchairmanm Just tagging you for accountability - updating this action to target a specific SHA as recommended to increase security

PR #2983, opened 2025-03-17
remove the note about webhook containers not being allowed in multi-main

netroy:
related https://github.com/n8n-io/n8n/pull/13989


netroy:
The core change has been released. Should we merge this @imchairmanm

PR #2990, opened 2025-03-17
Remove quote rule

evilnick:
It really is very annoying


imchairmanm (approved):
Thanks for the heads up on the change!.  You definitely don't need my approval for this one. I hate this one too and think it's objectively wrong :+1:

PR #2994, opened 2025-03-18
Review and update Docker Compose instructions

imchairmanm:
Made some updates to the Docker Compose instructions:

* Combined Docker and Docker Compose install sections since compose now gets pulled in as a dependency anyways
* Updated non-root access section to allow access to the Docker group without logging out or inputting your password again
* Reformatted DNS section
* Moved configuration sections up and put in dedicated directory
* Updated `docker-compose.yml` file:
    * Remove deprecated versioning
    * Added local directory to the compose file by default
    * Auto-create volumes with through compose instead of externally.

### DIFF
--- a/docs/hosting/installation/server-setups/docker-compose.md
+++ b/docs/hosting/installation/server-setups/docker-compose.md
@@ -16,7 +16,7 @@ You can find Docker Compose configurations for various architectures in the [n8n
 
 ## 1. Install Docker and Docker Compose
 
-How you install Docker and Docker Compose can vary depending on the Linux distribution you use. You can find detailed instructions in [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/) installation documentation. The following example is for Ubuntu:
+How you install Docker and Docker Compose can vary depending on the Linux distribution you use. You can find detailed instructions in both the [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/) installation documentation. The following example is for Ubuntu:
 
 ```bash
 # Remove incompatible or out of date Docker implementations if they exist
@@ -45,7 +45,7 @@ docker compose version
 
 ## 2. Optional: Non-root user access
 
-You can optionally grant access to run docker without the `sudo` command.
+You can optionally grant access to run Docker without the `sudo` command.
 
 To grant access to the user that you're currently logged in with (assuming they have `sudo` access), run:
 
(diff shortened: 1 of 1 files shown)


docs/hosting/installation/server-setups/docker-compose.md
```diff
@@ -1,70 +1,128 @@
...
 
 --8<-- "_snippets/self-hosting/warning.md"
 
 --8<-- "_snippets/self-hosting/installation/latest-next-version.md"
 
-### 1. Install Docker
+## 1. Install Docker and Docker Compose
 
-This can vary depending on the Linux distribution used. You can find detailed instructions in the [Docker documentation](https://docs.docker.com/engine/install/){:target=_blank .external-link}. The following example is for Ubuntu:
+How you install Docker and Docker Compose can vary depending on the Linux distribution you use. You can find detailed instructions in [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/) installation documentation. The following example is for Ubuntu:
```
evilnick, line 19:
```suggestion
How you install Docker and Docker Compose can vary depending on the Linux distribution you use. You can find detailed instructions in both the [Docker](https://docs.docker.com/engine/install/) and [Docker Compose](https://docs.docker.com/compose/install/) installation documentation. The following example is for Ubuntu:
```

docs/hosting/installation/server-setups/docker-compose.md
```diff
@@ -1,70 +1,128 @@
...
 
-### 3. Install Docker-Compose
+To grant access to a different user, type the following, substituting `<USER_TO_RUN_DOCKER>` with the appropriate username:
+
+```bash
+sudo usermod -aG docker <USER_TO_RUN_DOCKER>
+```
 
-This can vary depending on the Linux distribution used. You can find detailed instructions in the [Docker documentation](https://docs.docker.com/compose/){:target=_blank .external-link}.
+You will need to run `exec sg docker newgrp` from any of that user's existing sessions for it to access its new group permissions.
```
evilnick, line 64:
```suggestion
You will need to run `exec sg docker newgrp` from any of that user's existing sessions to access the new group permissions.
```
?

docs/hosting/installation/server-setups/docker-compose.md
```diff
@@ -1,70 +1,128 @@
...
+
+```bash
+docker --version
+docker compose version
 ```
 
-### 2. Optional: Non-root user access
+## 2. Optional: Non-root user access
+
+You can optionally grant access to run docker without the `sudo` command.
```
evilnick, line 48:
```suggestion
You can optionally grant access to run Docker without the `sudo` command.
```

evilnick (approved):
Just a few comments (I also added a description so we can hopefully get better indexing for this page).

docs/hosting/installation/server-setups/docker-compose.md
imchairmanm: Good catch! I think I removed some things from this sentence and forgot to read the whole thing again :smile:

PR #2998, opened 2025-03-18
Update npm.md to remove Windows note

StarfallProjects:
Remove note to Windows users - n8n will start without them navigating into that directory, and n8n should autodetect home folder.


imchairmanm:
Hey Debs! Good catch and thanks for the PR! :tada:

imchairmanm (approved):

PR #3000, opened 2025-03-19
Update templates API endpoints and document missing parameters

imchairmanm:


netroy (approved):
Thanks 🎉

netroy (approved):
Looks good.
The schemas are for individual items in the response object, but if someone complains, we can document the response objects as well then.

PR #3009, opened 2025-03-22
Add action to redirect support  requests

evilnick:
This action monitors the issue labels.
When an issue is labelled 'support' it will receive the automatic reply linking to the support forum and the issue will be closed.


imchairmanm (approved):
Oh, nice! This could definitely be helpful.

Issue #2975, opened 2025-03-16
[DOCS-ISSUE] - External link PR checker


evilnick:
**Where is the bug/mistake?**
The files being passed through to lychee should be filtered to be only .md files. Binary files break it, causing it to error out and issue a blank report

PR #3010, opened 2025-03-22
Update lychee-pr-check.yml

evilnick:
a slight re-write of the lychee-pr checker to replace the compromised action and harden against future issues by pinning external actions to specific versions

### DIFF
--- a/.github/workflows/lychee-pr-check.yml
+++ b/.github/workflows/lychee-pr-check.yml
@@ -12,20 +12,23 @@ jobs:
 
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
-      - name: Get changed docs files
-        id: changed-md-files
-        run: |
-          CHANGED_FILES=$(git diff --name-only $GITHUB_SHA^ $GITHUB_SHA -- '*.md'  | paste -sd "," -)
-          echo "Changed Markdown files: '$CHANGED_FILES'"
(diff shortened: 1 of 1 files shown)


evilnick:
Fixes #2975

.github/workflows/lychee-pr-check.yml
```diff
@@ -12,20 +12,23 @@ jobs:
...
-          echo "changed_files=${CHANGED_FILES:-}" >> "$GITHUB_OUTPUT"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - name: Get all changed markdown files
+        id: changed-markdown-files
+        uses: step-security/changed-files@3dbe17c78367e7d60f00d78ae6781a35be47b4a1 # v45.0.1
+        with:
+          # Avoid using single or double quotes for multiline patterns
+          files: |
+            **.md
+            docs/**.md
```
imchairmanm, line 23: Just out of curiousity, are files here not targetted by the glob above?

imchairmanm (approved):
Sounds like a good idea. Thanks for looking into this!

.github/workflows/lychee-pr-check.yml
evilnick: actually, I think both of these may be wrong depending on what globbing system is in use. It should probably be just:
docs/**/*.md

evilnick: the only real way to check is to merge it and make some PRs...

PR #3047, opened 2025-04-02
Update course level 2 to use pokeapi instead of quotable

elsmr:
quotable.io is no longer online it seems, this PR replaces it with PokeAPI in our level 2 course


Joffcom (approved):

imchairmanm:
Thanks @elsmr! This looks great.

PR #3052, opened 2025-04-05
remove http_proxy_license_server env var

csuermann:
Since all communication with license server always happens over https, the `http_proxy_license_server` variable will never be used and should be removed from the docs.


csuermann:
all good, @imchairmanm

PR #3086, opened 2025-04-16
docs: Remove tool name parameters from docs as they were removed

Cadiac:
Separate name parameter was removed from nodes

- Workflow Tool (v2.2)
- Vector Stores Tool (v1.1)
- Code Tool (v1.2)

and the name of the node is now used instead, to align all tool nodes to work in a similar way.

When the node name is used on the tool's description when passing it to the agent whitespaces on the name are converted to underscores - this is similar to how other tools already worked so perhaps we don't need to highlight this too much? Feel free to suggest something else for the docs here too.

These changes [were merged today](https://github.com/n8n-io/n8n/pull/14604), new versions of these nodes starting next release will not have the name parameter.

### DIFF
--- a/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
+++ b/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
@@ -19,17 +19,15 @@ For usage examples and templates to help you get started, refer to n8n's [Vector
 
 ## Node parameters
 
-### Data Name
-
-Enter the name of the data in the vector store.
-
 ### Description of Data
 
 Enter a description of the data in the vector store.
 
-n8n uses the **Data Name** and **Description of Data** parameters to populate the tool description for AI agents using the following format:
+n8n uses the **Node Name** and **Description of Data** parameters to populate the tool description for AI agents using the following format:
 
-> Useful for when you need to answer questions about [Data Name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
+> Useful for when you need to answer questions about [Node Name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
+
+Spaces in **Node Name** are converted to underscores on the tool description, and special characters are not allowed.
 
 ### Limit
 
--- a/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolcode.md
+++ b/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolcode.md
@@ -16,10 +16,6 @@ On this page, you'll find the node parameters for the Custom Code Tool node and 
 
 ## Node parameters
 
-### Name
-
-Give your custom code a name. It can't contain whitespace.
-
 ### Description
 
 Give your custom code a description. This tells the agent when to use this tool. For example:
--- a/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolworkflow.md
+++ b/docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolworkflow.md
@@ -16,10 +16,6 @@ On this page, you'll find the node parameters for the Workflow Tool node, and li
 
 ## Node parameters
 
-### Name
-
-Enter a name for your custom code. It can't contain whitespace or special characters.
-
 ### Description
 
 Enter a custom code a description. This tells the agent when to use this tool. For example:


docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
```diff
@@ -19,17 +19,15 @@ For usage examples and templates to help you get started, refer to n8n's [Vector
...
-### Data Name
-
-Enter the name of the data in the vector store.
-
 ### Description of Data
 
 Enter a description of the data in the vector store.
 
-n8n uses the **Data Name** and **Description of Data** parameters to populate the tool description for AI agents using the following format:
+n8n uses the **Node Name** and **Description of Data** parameters to populate the tool description for AI agents using the following format:
```
imchairmanm, line 26:
```suggestion
n8n uses the node name (select the name to edit) and **Description of Data** parameter to populate the tool description for AI agents using the following format:
```

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
```diff
@@ -19,17 +19,15 @@ For usage examples and templates to help you get started, refer to n8n's [Vector
...
+
+> Useful for when you need to answer questions about [Node Name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
```
imchairmanm, line 28:
```suggestion
> Useful for when you need to answer questions about [node name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
```

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
```diff
@@ -19,17 +19,15 @@ For usage examples and templates to help you get started, refer to n8n's [Vector
...
 
-> Useful for when you need to answer questions about [Data Name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
+Spaces in **Node Name** are converted to underscores on the tool description, and special characters are not allowed.
```
imchairmanm, line 30:
```suggestion
Spaces in the node name are converted to underscores in the tool description. Special characters aren't allowed.
```

Are special characters removed, do they break something, or does the field not accept them?

imchairmanm (approved):
Thanks for the PR @Cadiac! I left a few minor re-wording suggestions to satisfy Vale, our text linter, and to remove special formatting from the node name since it isn't specifically a parameter. I had one remaining question about special characters I hoped you could clarify for me.

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
Cadiac: The name field accepts the usual characters our nodes normally accept, but like on other tools you are unfortunately able to insert some special characters on the node names that langchain doesn't support.

Commonly spaces are something our users might enter on the tool name, and they are converted to underscores, so if user names the node "Test Name" its passed as `Test_Name` to the tool description that the agent sees.

If user names the tool with some other special characters they'll see an error when the agent runs.

imchairmanm: Okay cool. I was just wondering about what specifically happened when you attempted to add "not allowed" characters. I'll update it slightly then to mention that it'll cause problems for the agent. Thanks for clarifying!

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
```diff
@@ -19,17 +19,15 @@ For usage examples and templates to help you get started, refer to n8n's [Vector
...
 
 Enter a description of the data in the vector store.
 
-n8n uses the **Data Name** and **Description of Data** parameters to populate the tool description for AI agents using the following format:
+n8n uses the node name (select the name to edit) and **Description of Data** parameter to populate the tool description for AI agents using the following format:
+
+> Useful for when you need to answer questions about [node name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
 
-> Useful for when you need to answer questions about [Data Name]. Whenever you need information about [Description of Data], you should ALWAYS use this. Input should be a fully formed question.
+Spaces in the node name are converted to underscores in the tool description. Special characters aren't allowed.
```
imchairmanm, line 30:
```suggestion
Spaces in the node name are converted to underscores in the tool description. Special characters aren't allowed and will result in errors when the agent runs.
```

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolvectorstore.md
Cadiac: image

This is what happens if you have tools with names that contain characters that langchain doesn't support. Not optimal, and I suppose this is why the name field was once added on these couple of nodes, but they were removed now in favor of having every tool work the same.

PR #3089, opened 2025-04-19
Update broken blog URLs

mahyunyi:
Updated references from https://n8n.io/blog/ → https://blog.n8n.io/, which previously returned 404 errors.


imchairmanm (approved):
Thanks for the PR @mahyunyi! The blog recently was moved to a subdomain and had redirects that were working for some users and not for others. I don't think we would have been able to identify without reports like this. Appreciate it!

PR #3114, opened 2025-04-28
deprecate HTTP request tool and update HTTP request node

imchairmanm:


docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.toolhttprequest.md
```diff
@@ -7,6 +7,10 @@ contentType: [integration, reference]
 
 # HTTP Request Tool node
 
+/// warning | Stand-alone tool deprecated
+The stand alone HTTP Request tool is deprecated and its functionality is now incorporated in the standard [HTTP Request](/integrations/builtin/core-nodes/n8n-nodes-base.httprequest/index.md) node. You can now attach the standard HTTP Request node to AI agents to use it as a tool.
```
CharlieKolb, line 11: Would be nice if this is a bit more explicit about any possible required action. How about something like
```
New HTTP Request tools added to your workflows are now tool versions of the standard <HTTP Request link> node.

This page is about the legacy custom HTTP Request tool. You can differentiate them in your workflow by checking whether your node has an `Add Option` property - only the new version does.
```

CharlieKolb: Yeah from a user PoV it's basically just a node version change that happens to change a lot inside the node

imchairmanm: That's good to keep in mind from a framing perspective :+1:

With that being the case, would you suggest that we remove this from the nav (so that it's only accessible from the docs link from legacy nodes or do you think that'd cause more problems than it'd solve?

imchairmanm: Okay, reworked the note a bit and took the page out of the nav and search results. Any other feedback?

CharlieKolb (approved):
Looks great, thank you :D

PR #3118, opened 2025-04-29
Add common issue to MCP server about nginx config

imchairmanm:


jeanpaul (approved):
Awesome, looks good. Also kudos for explaining why you choose these options.
Might be interesting to know that the gzip compression is already in n8n itself, so that it's not really necessary to also specify that in your nginx config.

imchairmanm:
> Awesome, looks good. Also kudos for explaining why you choose these options. Might be interesting to know that the gzip compression is already in n8n itself, so that it's not really necessary to also specify that in your nginx config.

Okay, excellent! I added a mention of gzip from n8n. Thanks for the reviews!

PR #3137, opened 2025-05-07
Add info on custom scopes for microsoft entra id creds

imchairmanm:


imchairmanm:
@Joffcom I think you were mentioned on a ticket I de-duped for this one. If you wouldn't mind giving this a quick glance and approving if it looks alright, that'd be much appreciated.

PR #3149, opened 2025-05-09
Update configuration-methods.md

DevelopmentDePaulo:
In a lot of cases at using Windows in N8N self Hosting using NPM, command "export" doesnt exist, so needed to use instead "set" command to work.


StarfallProjects:
Thank you for the contribution - good call! 🎉I've made a tweak to expand it further.

PR #3154, opened 2025-05-12
adds PUT operation to variables endpoint of API

riascho:
This PR adds documentation for the new API endpoint for updating `variables` according to the [PR of the main n8n repo](https://github.com/n8n-io/n8n/pull/15315)

### DIFF
--- a/docs/api/v1/openapi.yml
+++ b/docs/api/v1/openapi.yml
@@ -952,6 +952,8 @@ paths:
         - Variables
       summary: Update a variable
       description: Update a variable from your instance.
+      parameters:
+        - $ref: '#/components/parameters/variableId'
       requestBody:
         description: Updated variable object.
         content:
--- a/docs/api/v1/openapi.yml
+++ b/docs/api/v1/openapi.yml
@@ -928,6 +928,23 @@ paths:
                 $ref: '#/components/schemas/variableList'
         '401':
           $ref: '#/components/responses/unauthorized'
+  /variables/{id}:
+    delete:
+      x-eov-operation-id: deleteVariable
+      x-eov-operation-handler: v1/handlers/variables/variables.handler
+      tags:
+        - Variables
+      summary: Delete a variable
+      description: Delete a variable from your instance.
+      parameters:
+        - $ref: '#/components/parameters/variableId'
+      responses:
+        '204':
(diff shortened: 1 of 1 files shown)


imchairmanm (changes requested):
Hey @riascho thanks so much for the PR! I had a look and I think that we might need to move this down a bit so it sits under the `/variables/{id}` section instead of the general `/variables` section. Can you take a look and see if that sounds right to you?

docs/api/v1/openapi.yml
```diff
@@ -928,6 +928,31 @@ paths:
...
+          description: Operation successful.
+        '400':
+            $ref: '#/components/responses/badRequest'
+        '401':
+          $ref: '#/components/responses/unauthorized'
+        '403':
+          $ref: '#/components/responses/forbidden'
+        '404':
+          $ref: '#/components/responses/notFound'
   /variables/{id}:
```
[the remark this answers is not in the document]
riascho: Ah yes you are right! Overlooked the `id` in the `yaml` [file](https://github.com/n8n-io/n8n/pull/15315/files#diff-ba4764e0cd57f6a1245d10640fa2376bb113897ce9d44e1318c7c457a7720992) 😅

docs/api/v1/openapi.yml
```diff
@@ -945,6 +945,31 @@ paths:
           $ref: '#/components/responses/unauthorized'
         '404':
           $ref: '#/components/responses/notFound'
+    put:
+      x-eov-operation-id: updateVariable
+      x-eov-operation-handler: v1/handlers/variables/variables.handler
+      tags:
+        - Variables
+      summary: Update a variable
+      description: Update a variable from your instance.
```
imchairmanm, line 954:
```suggestion
      description: Update a variable from your instance.
      parameters:
        - $ref: '#/components/parameters/variableId'
```

Just guessing on this one.

imchairmanm (approved):
Looks good Ria! I just left one further suggestion that I think puts the variable ID as a path parameter.

docs/api/v1/openapi.yml
riascho: So Sorry 😞 Totally overlooked that too! Thanks Justin! 🙏

PR #3160, opened 2025-05-15
Updated X Action documentation

ItRachii:
The X integration Action also supports to get the information of any user by id or username. It also has toggle to get the information of the user who's credentials are used to create the X connection.

### DIFF
--- a/docs/integrations/builtin/app-nodes/n8n-nodes-base.twitter.md
+++ b/docs/integrations/builtin/app-nodes/n8n-nodes-base.twitter.md
@@ -28,7 +28,8 @@ Refer to [X credentials](/integrations/builtin/credentials/twitter.md) for guida
     * Search tweets
     * Like a tweet
     * Retweet a tweet
-
+* User
+  	* Get a user
 ## Templates and examples
 
 <!-- see https://www.notion.so/n8n/Pull-in-templates-for-the-integrations-pages-37c716837b804d30a33b47475f6e3780 -->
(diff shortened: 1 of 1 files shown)


imchairmanm (changes requested):
Hey @ItRachii, thanks so much for the PR!

It looks like we might also be missing the "List" resource (it has one operation: add a user to a list). Would you mind also adding that to this list since you're already doing work here? Thanks!

ItRachii:
Hi @imchairmanm, Added the list action. Also, the documentation does not include any detailed information for the user so can I go ahead and provide a detailed documentation for X?

docs/integrations/builtin/app-nodes/n8n-nodes-base.twitter.md
```diff
@@ -28,7 +28,10 @@ Refer to [X credentials](/integrations/builtin/credentials/twitter.md) for guida
     * Search tweets
     * Like a tweet
     * Retweet a tweet
-
+* User
+  	* Get a user
+* List 
+    * Add a member to the list
```
imchairmanm, line 34:
```suggestion
    * Add a member to a list

```

imchairmanm (approved):
Looks good! Thanks for adding the additional resource and operation :+1:

imchairmanm:
> Hi @imchairmanm, Added the list action. Also, the documentation does not include any detailed information for the user so can I go ahead and provide a detailed documentation for X?

@ItRachii Yeah, this page is currently following a pattern we use in some places where we just list resources and their operations without discussing parameters and options in depth.

If we were going to add detailed information for the user, we'd want to add it for all of the resources, operations, and options. An example of a node that has this style of documentation is the [Gmail node](https://docs.n8n.io/integrations/builtin/app-nodes/n8n-nodes-base.gmail/). I think that might be a bit out of scope for this PR, but if you think it's important, feel free to submit a follow up that adds that information.

PR #3161, opened 2025-05-15
docs: compose.yaml and --api flag redundancy

blueprismo:
This PR makes two tiny changes:
1. Removes the `--api` flag, as `--api.insecure=true` implies it.
2. Changes the `docker-compose.yml` file definition to `compose.yaml` [ref](https://docs.docker.com/compose/intro/compose-application-model/#the-compose-file) is preferred since August 2020, v.1.27.0


imchairmanm (approved):
@blueprismo Thanks for the PR! These look like very reasonable changes.

PR #3163, opened 2025-05-15
Small verification guidelines improvement

pdwarf:
This PR adds two small refinements:
- clarification that verified community nodes are available on all deployment types, to make it clear they are also on cloud
- n8n reserves right to reject nodes that compete with paid features


imchairmanm (approved):
@pdwarf This looks good to me :+1: Is this ready to go? Just asking since it's marked as a draft currently.

PR #3181, opened 2025-05-22
Mention that Simple Vector Store data is accessible to all users

Cadiac:
Hey! We're changing Simple Vector Stores at PR https://github.com/n8n-io/n8n/pull/15421, changing the `Memory Key` picker to a resource locator that lists all currently existing vector stores in memory. This is introduced on a new version of the In-Memory vector store node, version 1.2, to be shipped at next n8n beta version once that PR gets merged.

Another substantial change at that PR is that we're removing the workflow prefixing done by n8n at the backend from these nodes on version 1.2. This removes the workflow level isolation and enables the Simple Vector Stores to be used across workflows, but this also means that any instance users can now access data stored on the in memory vector stores regardless of their access to workflows. We've adjusted the warning on the UI to mention this, but docs should also change here.

This docs page also mentions the workflow isolation bunch of times. I didn't touch those, but that is changing with the new version of the node. Old nodes still respect the workflow level isolation, but the new version can also select stores created by the older versions of the node by picking the workflow prefixed items from the list. Some adjustments to the texts here should be made.

### DIFF
--- a/docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.vectorstoreinmemory.md
+++ b/docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.vectorstoreinmemory.md
@@ -36,7 +36,7 @@ This node stores data in memory only. All data is lost when n8n restarts and may
 
 Memory keys for the Simple Vector Store node are global, not scoped to individual workflows.
 
-This means that all user of the instance can access vector store data by adding a Simple Vector Store node and selecting the memory key, regardless of the access controls set for the original workflow. Take care not to expose sensitive information when ingesting data with the Simple Vector Store node.
+This means that all users of the instance can access vector store data by adding a Simple Vector Store node and selecting the memory key, regardless of the access controls set for the original workflow. Take care not to expose sensitive information when ingesting data with the Simple Vector Store node.
 
 ## Node usage patterns
 


imchairmanm:
Hey @Cadiac thanks for the PR! So to summarize, the changes that the docs need to account for are:

* The simple vector store has a new interface for setting the memory key by selecting it from a list populated by with the memory keys the instance currently has available (I assume with a "create" option as one of the choices).
* Simple vector store data is now scoped to the instance instead of an idividual workflow. We don't have access controls in place, so users should be aware that the data inside will be accessible to any user of the instance.
* Workflow isolation with this node is no longer much of a thing and we should remove references to that throughout the docs.

Does that sound about right to you or do I need some adjustments?

Cadiac:
> Hey @Cadiac thanks for the PR! So to summarize, the changes that the docs need to account for are:
> 
> * The simple vector store has a new interface for setting the memory key by selecting it from a list populated by with the memory keys the instance currently has available (I assume with a "create" option as one of the choices).
> * Simple vector store data is now scoped to the instance instead of an idividual workflow. We don't have access controls in place, so users should be aware that the data inside will be accessible to any user of the instance.
> * Workflow isolation with this node is no longer much of a thing and we should remove references to that throughout the docs.
> 
> Does that sound about right to you or do I need some adjustments?

Exactly!

Old versions of SImple Vector Store will still behave as if they were workflow isolated, but other workflows using the newer version of the node can select vector stores created by these old nodes - they'll show up on the list with a workflow id prefix like `workflowid__memory-key-name-here`. I would remove the workflow isolation, or mention that its only on n8n <1.95

docs/integrations/builtin/cluster-nodes/root-nodes/n8n-nodes-langchain.vectorstoreinmemory.md
```diff
@@ -20,11 +20,24 @@ The simple vector storage described here is different to the AI memory nodes suc
...
 
+### Vector store data isn't persistent
+
+This node stores data in memory only. All data is lost when n8n restarts and may also be purged in low-memory conditions.
+
+### All instance users can access vector store data
+
+Memory keys for the Simple Vector Store node are global, not scoped to individual workflows.
+
+This means that all user of the instance can access vector store data by adding a Simple Vector Store node and selecting the memory key, regardless of the access controls set for the original workflow. Take care not to expose sensitive information when ingesting data with the Simple Vector Store node.
```
Cadiac, line 39: typo, `all users`?

Cadiac (commented):
Seems accurate, small typo to fix

imchairmanm:
> Seems accurate, small typo to fix

Ah, thanks for the catch. Should I go ahead and merge once fixed or do we need to wait on anything?

Cadiac:
I would say go ahead, this is already in beta and will be in stable next week

imchairmanm (approved):

PR #3188, opened 2025-05-25
Created Docs for the LLMs

LeonardoDaviti:
Created Script which is getting the whole docs as .md files and merges them into single file for use of LLMs. also merged all the examples into the single .json file to give LLMs knowledge how to create workflows for the n8n.
I know script and structure is little bit messy, but you can use it for future and let the developers download the whole docs as single file in single click.


WhiteTorn:
Updated Request, divided files from the folders, because initial merge was 1M Token, which is too much for the llms.

StarfallProjects:
@WhiteTorn thank you for this. I've added it to an internal issue where we're looking at a few ways to make the docs more AI integrated. Do you have any thoughts on the pros and cons of this approach versus providing an MCP? We already  use Kapa for our AI search, so we're evaluating [Build an MCP Server with kapa.ai](https://www.kapa.ai/blog/build-an-mcp-server-with-kapa-ai) as one approach.

I should note this work is in its early days, so may be a while before this gets resolved.

LeonardoDaviti:
I think docs would be better to evaluate and use locally, MCP may have some bugs, but it would be great to have both options

HumanistSerif:
@WhiteTorn We have added the ability to install n8n docs search directly into LLM-based workflows https://github.com/n8n-io/n8n-docs/pull/3982. So I'm closing this PR.

PR #3227, opened 2025-06-03
chore: Update public API docs

MarcL:
Updates the Public API documentation to reflect the changes to 
I didn't realise this wasn't automatic. :)

Fixes: [PAY-1852](https://linear.app/n8n/issue/PAY-1852/add-to-the-api-the-ability-to-provision-users-into-existing-projects)
Related: n8n-io/n8n#12329

### DIFF
--- a/docs/api/v1/openapi.yml
+++ b/docs/api/v1/openapi.yml
@@ -1074,7 +1074,7 @@ paths:
       tags:
         - Projects
       summary: Add one or more users to a project
-      description: Add one or more users to a project from your instance.
+      description: Add one or more users to a project on your instance.
       parameters:
         - name: projectId
           in: path
@@ -1124,7 +1124,7 @@ paths:
       tags:
         - Projects
       summary: Delete a user from a project
-      description: Delete a user from a project from your instance.
+      description: Delete a user from a project on your instance.
       parameters:
         - name: projectId
           in: path
@@ -1168,7 +1168,7 @@ paths:
           schema:
             type: string
       requestBody:
-        description: Updated project object.
+        description: Payload containing the new role to assign to the project user.
         content:
           application/json:
             schema:


docs/api/v1/openapi.yml
```diff
@@ -952,10 +952,8 @@ paths:
         - Variables
       summary: Update a variable
       description: Update a variable from your instance.
-      parameters:
-        - $ref: '#/components/parameters/variableId'
```
imchairmanm, line 956: Does it make sense to remove this? As far as I can tell, the variable ID is still required in the path.

docs/api/v1/openapi.yml
```diff
@@ -1022,7 +1020,12 @@ paths:
       summary: Delete a project
       description: Delete a project from your instance.
       parameters:
-        - $ref: '#/components/parameters/projectId'
+        - in: path
```
imchairmanm, line 1023: Just out of curiosity: this seems to render the same as before. Is defining these inline preferred now?

docs/api/v1/openapi.yml
```diff
@@ -1057,6 +1067,128 @@ paths:
...
         '404':
           $ref: '#/components/responses/notFound'
+  /projects/{projectId}/users:
+    post:
+      x-eov-operation-id: addUsersToProject
+      x-eov-operation-handler: v1/handlers/projects/projects.handler
+      tags:
+        - Projects
+      summary: Add one or more users to a project
+      description: Add one or more users to a project from your instance.
```
imchairmanm:
```suggestion
      description: Add one or more users to a project on your instance.
```

Maybe?

docs/api/v1/openapi.yml
```diff
@@ -1057,6 +1067,128 @@ paths:
...
+        '404':
+          $ref: '#/components/responses/notFound'
+  /projects/{projectId}/users/{userId}:
+    delete:
+      x-eov-operation-id: deleteUserFromProject
+      x-eov-operation-handler: v1/handlers/projects/projects.handler
+      tags:
+        - Projects
+      summary: Delete a user from a project
+      description: Delete a user from a project from your instance.
```
imchairmanm:
```suggestion
      description: Delete a user from a project on your instance.
```

docs/api/v1/openapi.yml
```diff
@@ -1057,6 +1067,128 @@ paths:
...
+          schema:
+            type: string
+        - name: userId
+          in: path
+          description: The ID of the user.
+          required: true
+          schema:
+            type: string
+      requestBody:
+        description: Updated project object.
```
imchairmanm:
```suggestion
        description: Payload containing the new role to assign to the project user.
```

Maybe? Just trying to match the style used in the `POST` action.

docs/api/v1/openapi.yml
```diff
@@ -1036,9 +1039,16 @@ paths:
       x-eov-operation-id: updateProject
       x-eov-operation-handler: v1/handlers/projects/projects.handler
       tags:
-        - Project
+        - Projects
```
imchairmanm, line 1042: Ah, thanks! I had noticed this the other day and forgot to fix!

imchairmanm (changes requested):
Thanks for the PR @MarcL! I left some suggestions, but feel free to ignore if they're not appropriate.

docs/api/v1/openapi.yml
MarcL: I'm not sure why this has been removed. My PR didn't touch variables so I wonder if we've just not copied across the generated YAML for a while?

This PR? https://github.com/n8n-io/n8n/pull/15315

docs/api/v1/openapi.yml
MarcL: I think this might have been the schema validator that suggested the change.

imchairmanm: Ah that makes sense.

PR #3232, opened 2025-06-03
feat(ado-3519)- docs for changes in document loader

nikhilkuria:
We are making a change the Data loader node with this task https://linear.app/n8n/issue/ADO-3519/default-text-splitting-in-the-data-loaders 
Currently, the user had to use a text splitter input node. We are making this easy by adding a new property in the data loader node. If the user chooses `simple`, we use a Recursive Character Text Splitter` under the hood. The user also has an option to pick the `custom` option and chose a text splitter of their choice. 
This affects two nodes 
- Default Data Loader
- GitHub Document Loader


imchairmanm:
Waiting until next minor release to merge.

PR #3233, opened 2025-06-04
Typo: 1s = 1000ms

gui-ying233:
10s = 10000ms


imchairmanm (approved):
Good catch! Thank for the PR.

PR #3246, opened 2025-06-06
docs: Add documentation for Reranker Cohere node

schrothbn:
Adds documentation around the reranking feature and the first node supporting it using cohere.

Related PRs:
https://github.com/n8n-io/n8n/pull/16051
https://github.com/n8n-io/n8n/pull/16014

### DIFF
--- a/docs/integrations/builtin/credentials/cohere.md
+++ b/docs/integrations/builtin/credentials/cohere.md
@@ -10,6 +10,7 @@ contentType: [integration, reference]
 You can use these credentials to authenticate the following nodes:
 
 * [Cohere](/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.lmcohere.md)
+* [Reranker Cohere](/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.rerankercohere.md)
 * [Embeddings Cohere](/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.embeddingscohere.md)
 
 ## Prerequisites


styles/config/vocabularies/default/accept.txt
```diff
@@ -199,6 +199,7 @@ Qualys
 Realtime
 Rekognition
 requester
+[Rr]erank(er|ing)?
```
imchairmanm:
```suggestion
[Rr]erank(er|ing|ed)?
```

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.rerankercohere.md
```diff
@@ -0,0 +1,31 @@
...
+# Reranker Cohere
+
+The Reranker Cohere node allows you to [rerank](/glossary.md#ai-reranking) the resulting chunks from a [vector store](/glossary.md#ai-vector-store). You can connect this node to a vector store.
+
+The reranker reorders the list of documents retrieved from a vector store for a given query in order of descending relevance.
+
+On this page, you'll find the node parameters for the Reranker Cohere node, and links to more resources.
+
+///  note  | Credentials
+You can find authentication information for this node [here](/integrations/builtin/credentials/cohere.md).
```
imchairmanm, line 17: I don't actually see a credential for Cohere in the linked PRs. Is there a Cohere credential? If not, we should remove this note.

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.rerankercohere.md
```diff
@@ -0,0 +1,31 @@
+---
+#https://www.notion.so/n8n/Frontmatter-432c2b8dff1f43d4b1c8d20075510fe4
+title: Reranker Cohere
```
imchairmanm, line 3: Is this node actually called "Reranker Cohere"? I would expect it to be called "Cohere Reranker" instead to follow the "<adjective> <noun>" construction that English typically uses. In one of the linked PRs, I see that the display name uses "Reranker Cohere" while the description uses "Cohere Reranker".

It'd be great to clear up which is the preferred name. I would recommend "Cohere Reranker" if all things are equal.

schrothbn: In general I'd agree that "Cohere Reranker" sounds better, but other Langchain sub-nodes follow the same naming pattern (see for example the "Embeddings <Vendor>" nodes), so Reranker Cohere matches that pattern.
The description should've also respected that I guess, that's an oversight there. 
For here and because the users will see the node in the editor as "Reranker Cohere" we should probably stick to that at least for now?

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.rerankercohere.md
schrothbn: The credentials existed already as we support cohere for embeddings and as LLM. The MRs add support for their rerank feature (and reranking in general) and we reuse the existing credentials. But the credentials page links to the nodes it's being used in so I should add this there as well. Thanks for the catch, I'll update the credentials page to link to the new node as well.

imchairmanm: Ah, I totally missed that we already had this, sorry! I think I was looking [here](https://github.com/n8n-io/n8n/tree/master/packages/nodes-base/credentials) instead of [here](https://github.com/n8n-io/n8n/tree/master/packages/%40n8n/nodes-langchain/credentials) (I should have just searched the docs probably).

docs/integrations/builtin/cluster-nodes/sub-nodes/n8n-nodes-langchain.rerankercohere.md
imchairmanm: Yeah, I guess we're inconsistent because for models, chat models, memory, and output parsers, etc., we're using the `<adjective> <noun>` construction I had expected (at least for display value - I imagine we use the "type" as a prefix with things like file names to make sorting easier).

Either way, the docs should follow the product naming. I was just questioning whether that was an oversight or intentional and whether it made sense to revisit. Thanks for the added context.

PR #3251, opened 2025-06-08
Edit REST Countries API request on Level Two

mahyunyi:
The REST Countries API recently required explicit specification of fields in the query to limit the response. 
This pull request updates the HTTP Request node instructions to include the query parameter 'fields' with the value 'name,region,subregion'. This ensures the request returns only the necessary country data for the tutorial.

For more information on the REST Countries API change: https://gitlab.com/restcountries/restcountries/-/issues/265


imchairmanm (approved):
Thanks for the PR @mahyunyi! It's great to know about the API changes and to update the tutorial accordingly.

PR #3252, opened 2025-06-08
Update chapter-2.md

jekidan:
Removed the information duplicated in the first point.


imchairmanm (approved):
Thanks for catching @jekidan and for the PR!

PR #3268, opened 2025-06-14
Fixed link at the end of the page

salomaoparkour:
Hyperlink should be pointing to Tips and Common issues. Instead it was pointing to Metric-base Evaluations.


imchairmanm (approved):
Thanks for catching and for the PR @salomaoparkour!

PR #3267, opened 2025-06-14
Update n8n-nodes-langchain.mcptrigger.md (change tab to space)

taeyeopkim1:
# Description
The place set to "tab" was displayed differently depending on the rendering environment, so it was changed to "space".

# Capture
https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-langchain.mcptrigger/#integrating-with-claude-desktop
image

https://github.com/n8n-io/n8n-docs/blob/main/docs/integrations/builtin/core-nodes/n8n-nodes-langchain.mcptrigger.md
image

https://github.com/n8n-io/n8n-docs/edit/main/docs/integrations/builtin/core-nodes/n8n-nodes-langchain.mcptrigger.md
image

PR #3287, opened 2025-06-20
Fix JSON formatting

autologie:
Fix JSON formatting by replacing tabs with spaces


taeyeopkim1:
same: https://github.com/n8n-io/n8n-docs/pull/3267

PR #3300, opened 2025-06-24
Remove queue mode warning

that-one-tom:
Evaluations can now be used in queue mode after https://github.com/n8n-io/n8n/pull/16307 has been released


imchairmanm (approved):
Thanks :+1:

PR #3301, opened 2025-06-24
Custom templates expect the N8N_TEMPLATES_ENABLED not to be set to false

nikhilkuria:


docs/workflows/templates.md
```diff
@@ -25,10 +25,6 @@ If you use n8n's template library, this takes you to browse [Workflows on the n8
 
 --8<-- "_snippets/workflows/templates/submit-templates.md"
```
nikhilkuria, line 27: https://github.com/n8n-io/n8n/blob/master/packages/frontend/editor-ui/src/components/MainSidebar.vue#L80

nikhilkuria: I think it's easier for the user to set only a single env variable (N8N_TEMPLATES_HOST) if they want to host their templates library

PR #3313, opened 2025-06-26
Update cloudflare.md

kmanwar89:
Update documentation to reflect API TOKEN phrasing, which is different than API KEYS


imchairmanm (approved):
Thanks for the PR @kmanwar89!

PR #3327, opened 2025-07-03
Update instructions for bypassing "unverified app" with Google OAuth

mieubrisse:
The current instructions don't work; I was already creating the OAuth application from the same user that I was authenticating with. To fix, I had to add my user to the list of "Test users".


_snippets/integrations/builtin/credentials/google/unverified-app.md
```diff
@@ -1,3 +1,3 @@
-If using the OAuth authentication method, you might see the warning **Google hasn't verified this app**. To avoid this, you can create OAuth credentials from the same account you want to authenticate. 
+If using the OAuth authentication method, you might see the warning **Google hasn't verified this app**. To avoid this, you can add your email to the list of testers for the app. To do so, go to the ["Audience" page](https://console.cloud.google.com/auth/audience) page and add the email you're signing in with to the list of "Test users".
```
StarfallProjects:
```suggestion
If using the OAuth authentication method, you might see the warning **Google hasn't verified this app**. To avoid this:

- If your app **User Type** is **Internal**, create OAuth credentials from the same account you want to authenticate.  
- If your app **User Type** is **External**, you can add your email to the list of testers for the app: go to the [**Audience** ](https://console.cloud.google.com/auth/audience) page and add the email you're signing in with to the list of **Test users**.
```

Issue #3290, opened 2025-06-23
[DOCS-ISSUE] - Image consistency in long tutorial


xibe:
**Where is the bug/mistake?**
https://docs.n8n.io/try-it-out/tutorial-first-workflow/#step-three-add-the-nasa-node-and-set-up-credentials

**What problem did you find?**
In Step 3.6.d (the "Add the NASA node and set up credentials" in the "Add the NASA node and set up credentials" section), the provided text is `{{ $today.minus(7, 'days') }}` which, in n8n, displays a result of `[DateTime: 2025-06-16T00:00:00.000-04:00]`.
This is not consistent with the screen capture that follows immediately, which uses `{{ $today.minus({days: 7}).toFormat('yyyy-MM-dd')}}
` and displays `2025-06-16`.

While this is super minor, it may bring confusion to beginners, who might feel that not having `2025-06-16` as a test result might break the rest of the tutorial, and they'd feel the need to type exactly what is in the screen capture, thus losing time -- and possibly patience.

**Can you suggest a fix?**
Two options:
* update the screen capture to use `{{ $today.minus(7, 'days') }}` and `[DateTime: 2025-06-16T00:00:00.000-04:00]`., 
* or indicate that both `{{ $today.minus(7, 'days') }}` and  `{{ $today.minus({days: 7}).toFormat('yyyy-MM-dd')}}` give similar results.

srinikshipit:
Can I pick this case?

StarfallProjects:
@LearnThoroughly sorry for the slow reply - yes, if you want to do this issue, that would be great!

StarfallProjects:
It looks like this one got done at some point, as the screenshot and text now match.

srinikshipit:
hi @StarfallProjects , thanks for your comment, apologies I did not reply to it!

PR #3374, opened 2025-07-18
Modify expression in tutorial-first-workflow to match the reference image

ninadnaik10:


ninadnaik10:
Fixes: issue #3290

imchairmanm:
Hey @ninadnaik10, thanks for the PR. You're right that the text and the image in the tutorial didn't match. However, it's the image that is currently out of date (the expression it uses no longer works). Can you update the image instead?

imchairmanm (approved):
Thanks @ninadnaik10!

PR #3401, opened 2025-07-24
remove unused cooke consent control

StarfallProjects:
As we're now using a cookie widget and no longer using the built-in cookie consent, remove the link to modify the built-in consent (it does nothing now)


StarfallProjects:
@imchairmanm likewise good catch. Done.

PR #3402, opened 2025-07-24
clear out old link syntax

StarfallProjects:
As we now have automatic external links opening in new tab, clear out all old syntax


imchairmanm:
@StarfallProjects So one of the downstream effects of removing this syntax is that Vale starts to check the lines that used to use it more strictly because the [exception here](https://github.com/n8n-io/n8n-docs/blob/5602bee3f31dfe0ff3f2ff21ad0389258b93f4e7/.vale.ini#L9) no longer applies. Do you think we should ignore those for now or try to fix as part of this?

StarfallProjects:
@imchairmanm I've removed that exception as no longer needed. 

I think Vale is failing here because we're asking it to lint too many files at once. 

The Vale extension in VS Code isn't flagging links for me (I think it should automatically ignore normal Markdown link syntax) so hopefully this is ok.

imchairmanm:
> @imchairmanm I've removed that exception as no longer needed.
> 
> I think Vale is failing here because we're asking it to lint too many files at once.
> 
> The Vale extension in VS Code isn't flagging links for me (I think it should automatically ignore normal Markdown link syntax) so hopefully this is ok.

Ah okay. I think when I initially experimented with this it was finding a bunch of spelling errors and broken links, but we've also done some work to catch those outside of this change, so maybe they're all resolved at this point. Thanks for checking and for updating the vale rule.

imchairmanm:
When I run Vale, it looks like the current docs have 152 errors across `docs` and `_snippets`. This PR has 249 across those same directories, so it looks like there are some new things that are failing.

That being said, I don't think that's a reason to not merge and I suspect most of them are just branding things that need to be added to the vocab list eventually. I think it probably makes sense to sort that out later on a best effort basis.

imchairmanm:
Hey @StarfallProjects, forgive me for thinking out loud here. I think my only remaining hesitation on merging this in right now is whether it'll make it challenging to pull these changes into the docs re-org branch. I'm periodically merging main into that and resolving the list of conflicts that arise from changes to files that don't exist or have been heavily modified in the new structure. I'm not sure how best to do that with a change set this large.

I might need to work on merging main into the re-org branch to get it up-to-date as of now, merge this PR, and then immediately merge main into the re-org branch again to minimize the drift and conflicts. I'll see if I can work on that after lunch so we can get this merged sooner rather than later.

imchairmanm:
Okay, the re-org branch is up-to-date, so I'm going to merge this one in. Thanks @StarfallProjects!

PR #3404, opened 2025-07-24
Fix lychee

StarfallProjects:
I'm seeing this error on some PRs when running lychee: `ENOENT: no such file or directory open 'lychee/out.md'`

I think this PR should fix it, but have to admit I'm not totally certain (in particular, I'm not sure if `output = "_lychee/output.md"` was correct) Do you @imchairmanm have more knowledge of lychee?


imchairmanm (approved):
@StarfallProjects Add this as another area I haven't played too much with. Nick set this up during his time here and it's mostly worked so I haven't played with it too much. Happy to see if this fixes. it though.

PR #3408, opened 2025-07-24
First pass on docs for workflow diffs view.

jcn8n:
@imchairmanm created a first pass on some docs for the new workflow diffs view we'll hopefully be releasing soon. Just a draft for now but maybe you could have a first pass on reviewing and see what you think about the structure.

### DIFF
--- a/docs/source-control-environments/using/compare-changes.md
+++ b/docs/source-control-environments/using/compare-changes.md
@@ -41,7 +41,7 @@ The diff view highlights three types of changes:
 
 ## Reviewing node changes
 
-For modified nodes, you can also look at the speciifc changes by doing the following:
+For modified nodes, you can also look at the specific changes by doing the following:
 
 1. Click on any modified node
 2. A JSON diff appears showing all changes for that specific node
--- a/docs/source-control-environments/using/compare-changes.md
+++ b/docs/source-control-environments/using/compare-changes.md
@@ -1,0 +1,71 @@
+---
+#https://www.notion.so/n8n/Frontmatter-432c2b8dff1f43d4b1c8d20075510fe4
+title: Compare changes with workflow diffs
+description: Use workflow diffs to compare local and remote changes 
+contentType: howto
+---
+
+# Compare changes with workflow diffs
+
+Workflow diffs allow you to visually compare changes between the workflow you have on an instance with what is currently in the a connected Git repository. This helps you understand exactly what has changed on workflow before you decide to either push or pull a workflow across different environments.
+
+/// note | Feature availability
+- Available on Enterprise
+- Workflow diffs are only available when the environments features has been enabled on an instance
+///
+
+## Accessing workflow diffs
+
+You can access workflow diffs from two locations:
+
+1. **When pushing changes**: Click the workflow diff icon in the commit modal alongside the workflow you want to review
+2. **When pulling changes**: Click the workflow diff in the modified changes modal alongside the workflow you want to review
+
+## Understanding the workflow diff view
+
+When you open a workflow diff, n8n displays two workflows vertically above and below:
+
+### When pushing
+* **Remote version in Git** (top panel): Latest version in your Git repository
+* **Local Instance version** (bottom panel): Current locally saved version of the workflow
+
+### When pulling  
+* **Local instance version** (top panel): Current version on your n8n instance
+* **Remote version in Git** (bottom panel): Version being pulled from the Git repository
+
+The diff view highlights three types of changes:
+
+* **Added nodes and connectors**: Any new node additions or connectors will show as green
+* **Modified nodes and connectors**: Any modifications to existing nodes or connectors will show as orange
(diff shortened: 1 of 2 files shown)


docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
...
+
+### When pushing
+* **Remote version in Git** (top panel): Latest version in your Git repository
+* **Local Instance version** (bottom panel): Current locally saved version of the workflow
+
+### When pulling  
+* **Local instance version** (top panel): Current version on your n8n instance
+* **Remote version in Git** (bottom panel): Version being pulled from the Git repository
+
+The diff view highlights three types of changes:
```
imchairmanm: It might be worth noting that these indicators are added to both workflow views (there's not an "original" and then one with the diffs).

docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
+---
+#https://www.notion.so/n8n/Frontmatter-432c2b8dff1f43d4b1c8d20075510fe4
+title: Compare changes with workflow diffs
+description: Use workflow diffs to compare local and remote changes 
+contentType: howto
+---
+
+# Compare changes with workflow diffs
+
+Workflow diffs allow you to visually compare changes between the workflow you have on an instance with what is currently in the a connected Git repository. This helps you understand exactly what has changed on workflow before you decide to either push or pull a workflow across different environments.
```
imchairmanm:
```suggestion
Workflow diffs allow you to visually compare changes between the workflow you have on an instance and the most recent version saved in your connected Git repository. This helps you understand the exact changes to the workflow before you decide to either push or pull it across different environments.
```

docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
...
+## Accessing workflow diffs
+
+You can access workflow diffs from two locations:
+
+1. **When pushing changes**: Click the workflow diff icon in the commit modal alongside the workflow you want to review
+2. **When pulling changes**: Click the workflow diff in the modified changes modal alongside the workflow you want to review
+
+## Understanding the workflow diff view
+
+When you open a workflow diff, n8n displays two workflows vertically above and below:
```
imchairmanm: It might be good to put the pattern for this into words. It looks like we'd want to tell users that in both cases, the workflow to be updated / overwritten is in the top panel, while the proposed replacement is in the bottom panel.

docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
...
+
+The diff view highlights three types of changes:
+
+* **Added nodes and connectors**: Any new node additions or connectors will show as green
+* **Modified nodes and connectors**: Any modifications to existing nodes or connectors will show as orange
+* **Deleted nodes and connectors**: Any nodes or connectors that have been deleted will show as red
+
+## Reviewing node changes
+
+For modified nodes, you can also look at the speciifc changes by doing the following:
```
imchairmanm:
```suggestion
For modified nodes, you can also look at the specific changes by doing the following:
```

docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. Click on any modified node
```
imchairmanm: This part might work better not as a list since only the first item seems to be a user action.

docs/source-control-environments/using/compare-changes.md
```diff
@@ -0,0 +1,71 @@
...
+2. A JSON diff appears showing all changes for that specific node
+3. This allows you to review exactly what configuration changes were made
+
+## Viewing summary of changes
+
+In the top right, you will see a "changes" button showing a count of the number of changes:
+
+1. Click the **Changes** button in the top right of the diff view
```
imchairmanm: Similar situation here.

docs/source-control-environments/using/compare-changes.md
jcn8n: @imchairmanm this depends a little bit on exactly what the change is. If its modified or added then it shows on both. But if it was a deletion for example, it would only show on the version it will eventually be deleted from, not the latest one its being push up with (If that makes sense). So technically there is an original and one with changes if I understand your point correctly.

docs/source-control-environments/using/compare-changes.md
jcn8n: Yes, good point and that's correct. I've added something in to make that clear. Let me know if this is what you had in mind

jcn8n:
@imchairmanm I've had another pass on this to get a basic version lined up. We're intending to release workflow diffs next week. Could you take another look to see if any other suggestions or improvements?

imchairmanm:
Holding off on merging until this is released.

PR #3444, opened 2025-08-06
Doc 1570 update theme

StarfallProjects:
- Updates Material Insiders to the latest version. Note that this now includes MkDocs 1.6.1, so have removed the standalone requirement for that. Going forward, we'll use the MkDocs version provided by the theme, to ensure compatibility.

- Removes the build-test GitHub action in favour of checking links as part of Netlify builds. Add `--strict` flag to the Netlify build command, and verified this worked (build failed when I broke a link, is passing otherwise) This removes a GitHub action and removes the need to maintain a separate test config file.

- Update mkdocs-macros-plugin to 1.3.7


imchairmanm:
@StarfallProjects So after looking at the link checking a bit, I think I can resolve all of the issues with anchor links so we can go ahead and increase the link validating strictness across the board. Do you mind if I push some changes here to work around the last few failure cases and then increasing the link validation strictness?

StarfallProjects:
@imchairmanm go for it.
When you say failure cases - the build is passing?

PR #3445, opened 2025-08-06
Only accessible with 127.0.0.1 (localhost)

engrjislam:
Only accessible with 127.0.0.1 (localhost). As a result, **127.0.0.1** is removed from

```
services:
  ...
  n8n:
    image: docker.n8n.io/n8nio/n8n
    ...
    ports:
      - "127.0.0.1:5678:5678"
  ...
```


StarfallProjects:
@engrjislam Thank you for the contribution!

The inclusion of localhost is deliberate, to limit the availability of the port. We only want n8n to be accessible to local apps and traefik (security best practice) So I'm closing this PR.

PR #3453, opened 2025-08-07
limit search to docs

StarfallProjects:


StarfallProjects:
I'm gonna merge this, but not convinced it solves the problem. It's laggy and it just shows page titles. The old site search was a better basic search.

PR #3473, opened 2025-08-13
Extand example to exclude nodes

ihortom:
It's not obvious for a regular user what the node name to use. The current example shows regular nodes names. LLM nodes have different reference. Replacing one of the regular nodes with a reference to the name of an LLM node might help to orient the reader in the right direction.


ivov:
@ihortom For docs review please ping the docs team. Thanks!

PR #3503, opened 2025-08-21
Update expressions.md

ixil:
Return the result of the IIFE - when generating a node output I found I required this.

# Debug info
- n8nVersion: 1.106.3
- platform: docker (self-hosted)
- nodeJsVersion: 22.17.0
- database: sqlite
- executionMode: regular
- concurrency: -1

### DIFF
--- a/docs/code/expressions.md
+++ b/docs/code/expressions.md
@@ -81,7 +81,7 @@ The following code use the Luxon date and time library to find the time between 
 
 
 ```js
-{{(()=>{
+{{return (()=>{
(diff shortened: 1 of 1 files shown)


StarfallProjects:
@ixil can you share the full example of where you found you needed the explicit `return`? On a quick test using the docs example, it doesn't seem needed:
image

StarfallProjects:
As the CLA is unsigned and there's no response to my question, I'm closing this for now. Feel free to reopen.

PR #3519, opened 2025-08-25
Update TheHive5 webhook version

Thibault-Van-Win:
The TheHive5 trigger in n8n now requires a webhook version of 1, as version 0 has been deprecated. Using version 0 causes the workflow to fail without an error message since the trigger expects the updated format
(https://github.com/n8n-io/n8n/blob/master/packages/nodes-base/nodes/TheHiveProject/TheHiveProjectTrigger.node.ts#L251).

This commit updates the documentation to reflect the correct version.


StarfallProjects (approved):
Thanks! 🎉

PR #3549, opened 2025-08-28
docs: add missing top bar menu options in Editor UI

Hasnainghori112:
Updated the Top Bar documentation to include the three-dots (⋮) menu.  This section now documents the ability to duplicate, rename, download workflows,  and import workflows from URL or file. This makes the description of the Editor UI  more complete for new users.


StarfallProjects:
Hi! Thanks for your contribution, but we are no longer updating the course. We expect to fully replace it in the coming months.

PR #3557, opened 2025-08-29
first pass at github action - spam only

StarfallProjects:


StarfallProjects:
Merging so I can test

PR #3593, opened 2025-09-01
docs: add chat node z.ai document

tomsun28:
hi, the related PR is https://github.com/n8n-io/n8n/pull/19044


tomsun28:
close now due https://github.com/n8n-io/n8n/pull/19044

PR #3632, opened 2025-09-08
chore: Update Form trigger node with new form elements

HumanistSerif:
* Added Checkboxes and Radio Buttons as new form elements
* Made small addition to Checkboxes to describe limits


imchairmanm (approved):
Hey, @HumanistSerif, looks good! :+1: At this point, either of us can merge this in to main and it'll get released. I'll let you do that this time. Once you do, after 3-5 minutes, you should be able to see the changes live.

imchairmanm:
@HumanistSerif A small thing: the CLA bot seems to be confused about who made the commit in this PR. I think maybe the email you used for your commit doesn't seem to be tied to your GitHub account all the way. You might want to double check which email you're using locally and what email addresses you have associated with your GH account to check that they match up.

PR #3653, opened 2025-09-12
Minor edit to CSS to clean up visual design

HumanistSerif:
* Added a specific CSS definition to improve visual design of left nav

### DIFF
--- a/docs/_extra/css/extra.css
+++ b/docs/_extra/css/extra.css
@@ -43,6 +43,11 @@
   --color-background-light: #FDF6EC;
 }
 
+/* This snippet targets the navigation nodes labled as data-md-level = 1 in the HTML.
+   These are the first-level headings in each section of the Docs site.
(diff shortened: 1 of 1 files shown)


docs/_extra/css/extra.css
```diff
@@ -10,16 +10,43 @@
...
-
   --color-primary: #530092;
   --color-text-base: #58585F;
   --color-text-light: #FDF6EC;
   --color-background-dark: #040506;
   --color-background-light: #FDF6EC;
+}
 
-  
+nav[data-md-level="1"] > .md-nav__list > .md-nav__item {
```
StarfallProjects, line 51: Could you add a comment above it to explain what it does? I know we don't have consistent conventions in this file, so it's not obvious to do.

HumanistSerif: Great flag @StarfallProjects. I have added a comment to explain how this snippet works.

PR #3654, opened 2025-09-12
fix docker compose commands

gml3ff:
Docker compose command here is incorrect, should be `docker-compose` instead

image


StarfallProjects:
@gml3ff thanks for raising this! However, we're using the current Docker syntax here. The modern approach is `docker compose`, as shown in the official docs: https://docs.docker.com/reference/cli/docker/compose/

You might be thinking of the older standalone `docker-compose` tool, or possibly have an alias set up on your system?

Since our guide follows the current Docker documentation, I'll close this issue.

PR #3659, opened 2025-09-15
docs: clarify execution.resumeUrl not applicable for MCP Server Trigg…

jvinodraj:
- Added a new section "Not applicable in MCP" under execution reference.
- Explained that MCP does not expose a webhook-based resume URL for workflow continuation.
- Clarified that MCP sessions are stateful but differ from n8n’s execution-based state.


StarfallProjects:
@jvinodraj thanks for raising this 🎉 I've made an edit, as it's not only the MCP Server Trigger affected, and it makes sense to group this info with the relevant variable.

I'm also looking into getting a list of nodes that do/don't support `$execution.resumeUrl`, but in the mean time, let's get this merged. Thank you for flagging the problem!

StarfallProjects (approved):

jvinodraj:
Thanks for polishing and merging this! Good call on making it broader than just the MCP Server Trigger.

PR #3666, opened 2025-09-17
docs: Fix task runner python variables table syntax error

Cadiac:
The table was formatted wrong, leading into a broken docs page where the table wasn't rendered.


imchairmanm (approved):
Thanks for catching and for the PR @Cadiac :+1:

PR #3672, opened 2025-09-18
add Windows file locations for custom nodes

### LINKED ISSUE #2606: Documentation Clarification Request
URL [creating-nodes/build/node-development-environment](https://docs.n8n.io/integrations/creating-nodes/build/node-development-environment/)

Request and Reasoning:
Clarification of locations of key files when installed on windows, mac and linux, for the runtime of the dev environment.
In Windows, when the global n8n launches for local development, the place for putting your custom nodes is actually found in the current user directory, not in the node modules directory itself. I may not be the sharpest knife in the kitchen when it comes to the node ecosystem, so this may be a hard case of user error during setup of the devkit, but this one stumped me for a day or two of googling until I found the answer on a form thread. For past me, I would appreciate it if there could be a note added about where the runtime config is found on windows, mac and linux.

Example: after running n8n, you can add your custom directory for installing nodes if it does exist and link your build. On windows it be found under `%userprofile%/.n8n` in windows explorer, or `~/.n8n` in powershell.

Thanks for your consideration.

StarfallProjects:
Resolves https://github.com/n8n-io/n8n-docs/issues/2606

### DIFF
--- a/_snippets/integrations/creating-nodes/testing.md
+++ b/_snippets/integrations/creating-nodes/testing.md
@@ -27,8 +27,8 @@ You can test your node as you build it by running it in a local n8n instance.
 
     Windows:
 
-    - `C:\Users\<your-username>\.n8n\nodes`
-    - `C:\Users\<your-username>\.n8n\<your-custom-name>`
+    - `C:\Users\<user-profile>\.n8n\nodes`
+    - `C:\Users\<user-profile>\.n8n\<your-custom-name>`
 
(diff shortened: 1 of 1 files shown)


StarfallProjects:
Concerned about this one, as `custom` (which should be the default custom directory) didn't work on my Windows machine.

Joffcom:
Closing for now as I missed this and we are moving to GitBook

PR #3726, opened 2025-09-30
Document 'Create a task' operation for Asana

ChobPT:
Added documentation for the 'Create a task' operation in Asana integration, including field descriptions and output schema.


HumanistSerif:
@ChobPT we are working on automating the generation of node docs so this expansion for only one action seems redundant.

ChobPT:
You're welcome

PR #3790, opened 2025-10-21
Adding data migration tool to 1.116.0 release notes

prutwo:
just adding some text to the 1.116.0 release note


Joffcom:
@prutwo try following the same format we used for new features in 1.115 where we used some kind of callout box

prutwo:
@Joffcom 
I was wondering about that ... I saw some release notes are in callout box and some are not.
Adding it now.

PR #3799, opened 2025-10-24
docs: include example for separate runner pod in K8s

lindner-tj:
I couldn't get the task runner working using the sidecar method, but a dedicated runner Pod works, so I thought it was a good idea to include in the docs.


RoRoJ:
Hi [lindner-tj](https://github.com/lindner-tj), thank you for your contribution. We don't currently document k8s deployment options, so I'll close this for now, but bear it in mind if we extend our coverage in the future.

PR #3829, opened 2025-10-30
add documentation around credential override persistence

thekar22:
> [!NOTE]
> Documents database persistence for credential overwrites and adds a link to the full environment variables index.
> 
> - **Docs**
>   - **Embed configuration (`docs/embed/configuration.md`)**:
>     - Add *Persistence* section explaining `CREDENTIALS_OVERWRITE_PERSISTENCE`, database storage of encrypted overwrites in `settings`, and broadcast of `reload-overwrite-credentials` to workers.
>   - **Environment variables — credentials (`docs/hosting/configuration/environment-variables/credentials.md`)**:
>     - Add `CREDENTIALS_OVERWRITE_PERSISTENCE` variable with default `false` and description for multi-instance/queue propagation.
>   - **Environment variables overview (`docs/hosting/configuration/environment-variables/index.md`)**:
>     - Add link to the full environment variables reference index.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d6bb8cd3a83804de8173d80d0687aeb33d0a2428. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


docs/hosting/configuration/environment-variables/credentials.md
```diff
@@ -19,4 +19,5 @@ Enable credential overwrites using the following environment variables. Refer to
 | :------- | :---- | :------- | :---------- |
 | `CREDENTIALS_OVERWRITE_DATA`<br>/`_FILE` | * | - | Overwrites for credentials. |
 | `CREDENTIALS_OVERWRITE_ENDPOINT` | String | - | The API endpoint to fetch credentials. |
+| `CREDENTIALS_OVERWRITE_PERSISTENCE` | Boolean | `false` | Enable database persistence for credential overwrites. Required for multi-instance/queue mode to propagate overwrites to workers via Pub/Sub. |
```
rehevkor5: Not just queue mode though right? Impacts the behavior no matter how you run it.

thekar22: I can see the argument, but keeping the secret in memory may be sufficient for single instance setups. Whereas, in multi-instance configurations, it feels more like a requirement than a suggestion.

rehevkor5: It seems like the situation would be even worse in single-instance setups. The info would always get lost if the process ever crashes, dies, or stops in any way. At least if you have multiple instances you could cluster them so that - even if it's only in-memory - there's a better chance of at least one always being alive and thereby not losing the info - though I don't know if n8n does that or not.

thekar22: You're right that some instances in a multi-instance setup will keep the values in memory. That said, there would be no current way to re-propagate to the instance that crashed.

For the single instance setup, there are certainly consequences to deal with in the event of a crash. However, I would pose that this is still not a requirement. In this scenario, the instance does properly function until a crash (whereas, there is explicitly no propagation occurring for multi-instance) and the repopulation of credential overrides could be handled by a new API call.

HumanistSerif (approved):

PR #3933, opened 2025-11-26
Updated LangSmith setup instructions 

### LINKED ISSUE #3908: [DOCS-ISSUE] - LangSmith integration env vars
> **NOTE**: This is for issues with _documentation_ of n8n. If the problem is actually a bug in the software, [file a bug here instead](https://github.com/n8n-io/n8n/issues). If you need help, [ask on the forum](https://community.n8n.io/c/questions/12).

## Where is the bug/mistake?

https://docs.n8n.io/advanced-ai/langchain/langsmith/

## What problem did you find?
Missing LANGCHAIN_PROJECT optional env var to the table.

During the initial setup in LangSmith there is a project already there that is named "default". But it is hidden in the UI before the first trace collected, so it's hard to understand what's going on.
It would be great to have LANGCHAIN_PROJECT in the table, say that it's optional and will have a value of "default" if not set. Also it would be great to write a yellow note that "If you just created your LangSmith account, then you will see the project named "default" and LLM traces inside when the first trace is sent from n8n." 

As a side note there can be a lag between collecting and sending traces due to LANGCHAIN_CALLBACKS_BACKGROUND env var, which default value I don't clearly understand.

## Can you suggest a fix?
Add all available env vars from 
https://docs.langchain.com/langsmith/trace-with-langchain#1-configure-your-environment  
and 
https://docs.langchain.com/langsmith/trace-with-langchain#log-to-a-specific-project

Check if they should be named LANGSMITH_*, not LANGCHAIN_* like it is right now, depends on the SDK version you use.

## Additional context

none.

akshayamadhuri:
Added additional environment variables for LangSmith configuration.
## Summary

Adds missing `LANGCHAIN_PROJECT` and `LANGCHAIN_CALLBACKS_BACKGROUND` environment variables to the LangSmith documentation table, along with a note explaining the "default" project behavior.

## Changes

- Added `LANGCHAIN_PROJECT` env var (optional, default: `default`)
- Added `LANGCHAIN_CALLBACKS_BACKGROUND` env var (optional, default: `true`)
- Added "Required" and "Default Value" columns to the table
- Added note explaining "default" project visibility

## Fixes

#3908


v-karbovnichy:
Looks great!

PR #4216, opened 2026-02-11
docs: Add use Opik with n8n guide

vincentkoc:
## Summary
- add a new how-to: `docs/advanced-ai/opik.md`
- document self-hosted setup using `n8n-observability`

### DIFF
--- a/docs/advanced-ai/opik.md
+++ b/docs/advanced-ai/opik.md
@@ -99,6 +99,8 @@ volumes:
 services:
   n8n:
     build: .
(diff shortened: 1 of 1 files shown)


HumanistSerif:
Hi @vincentkoc. Thanks for the submission. I see that this workflow is already documented on the Opik Docs site. I'm going to close this PR in favour of pointing users to the Opik site. This way, we can ensure that the content isn't duplicated and doesn't need periodic syncing between two sources.

PR #4217, opened 2026-02-11
docs: clarify how expressions are evaluated in credential fields

### LINKED ISSUE #4149: [DOCS-REQUEST] - Credential Expression evaluation
## What information is missing?
It looks like there is no official documentation clarifying how n8n evaluates credentials when using it along with [expression](https://docs.n8n.io/code/expressions). 

For example, suppose a webhook node receives an access token, and that token is referenced in the credentials of a subsequent node using an expression like the one below. 

`{{ $('Webhook').item.json.headers.authorization }}`

Please find below the screenshot. Here, n8n appears to evaluate the credential using the execution data from the corresponding workflow execution. 

However, without this being documented, it's difficult for a user to interpret whether the dynamic evaluation of the credential is always scoped to the individual workflow execution and multiple workflow executions do not interfere with each other.

## Describe the page/pages you'd like

The section below explains how to use the expression to retrieve webhook data in the subsequent node. 
https://docs.n8n.io/code/expressions/#example-get-data-from-webhook-body

A documentation about the scope of expression evaluation, along with its usage with Credential, would be greatly beneficial in this section.

## Additional context

Image

MoonBrillante:
Closes #4149

This PR adds a section to the Expressions documentation clarifying how expressions are evaluated when used in credential fields.

It explains that expressions are evaluated within the current workflow execution context and that executions do not share data.

---
## Summary by cubic
Clarifies how expressions are evaluated in credential fields: they run within the current workflow execution and can access data from previous nodes. Each execution is isolated, so credential expressions don’t share data across runs; includes a brief webhook token example.

<sup>Written for commit e5948985a3fcb3333cea5b5d1528fb4b2d87eb81. Summary will update on new commits.</sup>

### DIFF
--- a/docs/code/expressions.md
+++ b/docs/code/expressions.md
@@ -70,6 +70,17 @@ This expression:
 1. Accesses the incoming JSON-formatted data using n8n's custom `$json` variable.
 2. Finds the value of `city` (in this example, "New York"). Note that this example uses JMESPath syntax to query the JSON data. You can also write this expression as `{{$json['body']['city']}}`.
 
+### Using expressions in credentials
+
+You can also use expressions in credential fields. When you reference data using expressions (for example, `{{$json.body.city}}` or `{{ $('Webhook').item.json.headers.authorization }}`), n8n evaluates the expression within the context of the current workflow execution.
+
+This means that:
+
+- Expressions in credentials can access data available in the current execution context, including data from previous nodes.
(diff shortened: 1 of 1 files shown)


HumanistSerif:
Thanks for this contribution, @MoonBrillante. We recently revamped the section on Working with Data and the file that this PR edits is in a different location. I moved your changes over to the new file in #4284.

PR #4386, opened 2026-03-21
fix(docs): simplify NASA API credential setup to use DEMO_KEY

stakiran:
Replace the multi-step API key generation process with DEMO_KEY, which works without registration and is sufficient for the tutorial.

For details, see [NASA Open APIs](https://api.nasa.gov/).

---
## Summary by cubic
Simplified the NASA API setup in the first workflow tutorial by using `DEMO_KEY` instead of a registered key. This removes email and registration steps so readers can complete the tutorial faster.

<sup>Written for commit 4132a05fec82a23b54d5e47a405d4ec958612dfa. Summary will update on new commits.</sup>


HumanistSerif:
@stakiran, thanks for your contribution. While you can use `DEMO_KEY` for API calls with small payloads, NASA recommends creating a proper API key using your email address for rate controlled API calls. This process, could admittedly, take a few more minutes to complete, but is more representative of how credentials are created for most services in real-world scenarios.

I'm therefore going to close this PR.

PR #4487, opened 2026-04-23
docs: add deprecated and versioned nodes reference page DOCS-956

its-imad:
Adds a reference page listing removed, deprecated, and versioned nodes across nodes-base and nodes-langchain. 
Includes nav entry under Built-in nodes and iCalendar added to Vale vocabulary.

Linear ticket: DOC-956

https://linear.app/n8n/issue/DOC-956/node-graveyard

### DIFF
--- a/docs/integrations/builtin/deprecated-and-versioned-nodes.md
+++ b/docs/integrations/builtin/deprecated-and-versioned-nodes.md
@@ -1,0 +1,204 @@
+---
+contentType: reference
+---
+
+# Deprecated and versioned nodes
+
+n8n improves its node library over time. This page lists removed nodes (fully removed), deprecated nodes (retired but still functional), and versioned nodes (active with multiple versions available).
+
+## Deprecated nodes
+
+n8n won't release further updates or bug fixes for deprecated nodes. Existing workflows that use them continue to run, but you should migrate to a supported alternative.
+
+/// note | Migrate deprecated nodes
+Replace deprecated nodes in your workflows before n8n removes them in a future release.
+///
+
+| Node | Final node version |
+|------|:-----------------:|
+| Binary Input Loader | 1 |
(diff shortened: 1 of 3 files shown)


docs/integrations/builtin/deprecated-and-versioned-nodes.md
```diff
@@ -0,0 +1,204 @@
...
+contentType: reference
+---
+
+# Deprecated and versioned nodes
+
+n8n improves its node library over time. This page lists removed nodes (fully removed), deprecated nodes (retired but still functional), and versioned nodes (active with multiple versions available).
+
+## Deprecated nodes
+
+n8n won't release further updates or bug fixes for deprecated nodes. Existing workflows that use them continue to run, but you should migrate to a supported alternative.
```
HumanistSerif: Please replace with "Existing workflows that use them will continue to run..."

RoRoJ (approved):
This looks good to me, just a few thoughts:

1. Maintenance - How will we know to update the page when a node is deprecated/removed/up-versioned? Will the automation be triggered automatically by a pertinent change in the code base?

2. Longer term question -Should we have an admonition on each versioned node's documentation page that states it's a versioned node, and links back to this page?

its-imad:
1. for now it will be a simple time trigger check once a week… get the current page analyse the code base generate the same page and check the diff, if diff create a new page , ask validation via slack, create linear ticket > branch >  PR . Tomorrow the "drift catcher" agent will catch new pr and check if there is a drift against the doc

its-imad:
2. I think it's an excellent idea as for now the info is hidden in the release notes sometimes and the code

RoRoJ:
> 2. I think it's an excellent idea as for now the info is hidden in the release notes sometimes and the code

Great, I'll add it to the list of stuff we want the node doc generator automation to do

PR #4520, opened 2026-04-27
docs: add MiniMax Chat Model node and credentials pages

its-imad:
## Summary

- Adds MiniMax Chat Model node documentation page
- Adds MiniMax credentials documentation page
- Adds both entries to nav.yml in alphabetical position (node between Lemonade and Mistral, credentials between mist and mistral)

## Linear ticket

https://linear.app/n8n/issue/DOC-1811/docs-for-minimax-ai-agent-model-node


RoRoJ (approved):
There's the question of whether we want to include parameter names (as distinct from their display names) and types/defaults in the bid to be more agent-friendly, currently the node doc generator is including this info. But I think since we haven't finalised the template yet, we can worry abotu it later.

its-imad:
Thanks Rowena! I didn't use the node doc generator this time,wanted to go through the process manually at least once to understand it. Will use it going forward. Agreed I'll keep this page aligned with the current template for now and revisit when the agent-friendly format is finalised. Happy to update it in a follow-up once everything is stable

PR #4532, opened 2026-04-28
feat: Add docs for workflow level OTEL

geemanjs:


docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,210 @@
...
+
+# OpenTelemetry tracing
+/// warning | This feature is still under development
+Agent telemetry and otel formatted metrics will be coming soon 
+///
+
+n8n can emit [OpenTelemetry](https://opentelemetry.io/){:target="_blank" .external-link} traces for workflow and node executions. Use these traces to monitor execution latency, debug failures, and track requests across services in your observability stack.
+
+/// info | Feature availability
+OpenTelemetry workflow tracing requires environment variables to be set so is currently available on self-hosted n8n only.
```
konstantintieber: Let's also include "from version 2.19.0" here

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,210 @@
...
+n8n also handles trace context propagation:
+
+- **Inbound**: If a webhook request includes a [W3C `traceparent` header](https://www.w3.org/TR/trace-context/){:target="_blank" .external-link}, n8n uses it as the parent for the workflow span. This links the trace to the upstream caller.
+- **Outbound**: HTTP Request nodes (and other nodes that use the n8n HTTP helpers) can inject a `traceparent` header into outbound requests. Downstream services that support W3C trace context can therefore continue the trace.
+- **Sub-workflows**: A sub-workflow's span uses the parent workflow's span as its parent.
+- **Resumed workflows**: When a workflow resumes after a wait, the new span links back to the previous span using a span link.
+
+## Enable tracing
+
+Set the following environment variables on each n8n instance you want workflow tracing enabled (main, workers, and webhook processors):
```
konstantintieber, line 41: Should there be a "where you want workflow tracing enabled"?

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,210 @@
...
+
+n8n appends `/v1/traces` to the endpoint by default. Point `N8N_OTEL_EXPORTER_OTLP_ENDPOINT` at the base URL of your collector, not the traces path.
+
+If your collector needs authentication, set `N8N_OTEL_EXPORTER_OTLP_HEADERS` to a comma-separated list of `key=value` pairs:
+
+```bash
+export N8N_OTEL_EXPORTER_OTLP_HEADERS="authorization=Bearer <your-token>,x-tenant=acme"
+
+// It is recommendded to use the `_FILE` postfix if you are putting a token here:
+N8N_OTEL_EXPORTER_OTLP_ENDPOINT_FILE=/mnt/otel-headers
```
konstantintieber: Should have an export in front as well?

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+n8n prefixes each key with `n8n.node.custom.` on the exported span. Values must be strings, numbers, or booleans.
+
+This API isn't available from the Code node. It's intended for node authors who want to enrich spans with domain-specific data.
+
+## Try it out with Jaeger
+
+You can send traces to a local [Jaeger](https://www.jaegertracing.io/){:target="_blank" .external-link} instance to see them in action.
```
konstantintieber, line 120: I love seeing this section here. I do worry about the fact that many readers of these docs may not know how to run n8n locally in the first place. But when thinking about how to describe "In order to run n8n locally,... " in here, it feels like too much scope creep on the page.
I just hope that those who do want to try it out with a temporary local n8n instance, find their way to https://github.com/n8n-io/n8n/blob/master/CONTRIBUTING.md with the setup and dev commands

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,210 @@
...
+
+3. Start n8n with tracing turned on and pointed at Jaeger:
+
+```bash
+export N8N_OTEL_ENABLED=true
+export N8N_OTEL_EXPORTER_OTLP_ENDPOINT=http://127.0.0.1:4318
+n8n start
+```
+
+5. Run a workflow, then open the Jaeger UI at [http://localhost:16686](http://localhost:16686) to see the trace.
```
konstantintieber: Since we may be talking to an audience that has never used jaeger before, I think it'd be worth to add these two extra steps:

> 'Select "n8n" as service" and click "Find traces" to see the OpenTelemetry traces emitted by n8n.

docs/hosting/configuration/environment-variables/opentelemetry.md
```diff
@@ -15,14 +15,14 @@ hide:
...
-| `N8N_OTEL_EXPORTER_SERVICE_NAME` | String | `n8n` | Value of the `service.name` resource attribute on exported spans. |
-| `N8N_OTEL_TRACES_SAMPLE_RATE` | Number | `1.0` | Fraction of traces to export, between `0` and `1`. n8n uses a trace ID ratio sampler, so all spans in a trace are either sampled or dropped together. |
-| `N8N_OTEL_TRACES_INCLUDE_NODE_SPANS` | Boolean | `true` | Whether to emit a `node.execute` span for each node execution. Set to `false` to export workflow-level spans only. |
-| `N8N_OTEL_TRACES_INJECT_OUTBOUND` | Boolean | `true` | Whether to inject the W3C `traceparent`/`tracesstate` headers into outbound HTTP requests made by nodes that use n8n's HTTP helpers. |
-| `N8N_OTEL_STARTUP_CONNECTIVITY_TIMEOUT_MS` | Number | `2000` | Timeout (in milliseconds) for the startup connectivity check against the OTLP endpoint. n8n logs a warning if the endpoint isn't reachable within this period. |
+| Variable | Type | Default | Description | Available from | 
+| :------- | :--- | :------ | :---------- | :------
```
konstantintieber, line 19: Repeating 2.19.0 per row in this table seems redundant to me.
Especially as long as it's all the same version for each one.
What do you think about adding a single note in the top that this is available from 2.19.0?

konstantintieber: See https://docs.n8n.io/external-secrets/#:~:text=vault%20per%20provider.-,From%20version%202.13.0,-%2C%20if%20enabled%2C%20project

geemanjs: We are about to add some new env vars so I feel this will be useful then.

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,212 @@
+---
+title: OpenTelemetry tracing
+description: Send workflow and node execution traces from n8n to an OpenTelemetry collector.
+contentType: howto
+---
+
+# OpenTelemetry tracing
+/// warning | This feature is still under development - Initially available from 2.19.0
```
konstantintieber, line 8: I think the heading is easier to miss, than having the "only available from 2.19.0" inside the notice.

Image

I would suggest to put "OTel is only available from version 2.19.0" into the notice.

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,212 @@
...
+export N8N_OTEL_EXPORTER_OTLP_HEADERS="authorization=Bearer <your-token>,x-tenant=acme"
+
+// For added protection - It is recommended to use the `_FILE` postfix if you are putting a token in here:
+export N8N_OTEL_EXPORTER_OTLP_HEADERS_FILE=/mnt/otel-headers
+```
+
+For the full list of supported variables, refer to [OpenTelemetry environment variables](/hosting/configuration/environment-variables/opentelemetry.md).
+
+/// note | Queue mode
+In [queue mode](/hosting/scaling/queue-mode.md), set the OpenTelemetry variables on the main, worker, and webhook instances. Trace context is propagated between main, worker and webhook instances.
```
konstantintieber, line 64: Basically they need to set these env vars on all n8n instances, right?
You spelled them out as "main,worker and webhook", but putting myself in the shoes of a user I'd be thinking "am I missing some other type of instance?".
May be worth to note that it should be all of them.

docs/hosting/logging-monitoring/opentelemetry.md
```diff
@@ -0,0 +1,212 @@
...
+
+# OpenTelemetry tracing
+/// warning | This feature is still under development - Initially available from 2.19.0
+Agent telemetry and open telemetry formatted metrics will be coming soon 
+///
+
+n8n can emit [OpenTelemetry](https://opentelemetry.io/){:target="_blank" .external-link} traces for workflow and node executions. Use these traces to monitor execution latency, debug failures, and track requests across services in your observability stack.
+
+/// info | Feature availability
+OpenTelemetry workflow tracing requires environment variables are set so is available on self-hosted n8n only.
```
sandra0503, line 15: Could be more simple?
```
OpenTelemetry workflow tracing is only available on self-hosted n8n.
```

konstantintieber (approved):
Def good enough to ship.
Improvements can be considered for follow-ups

docs/hosting/logging-monitoring/opentelemetry.md
konstantintieber: @geemanjs how do you feel about moving that version note down into the note from the heading?

PR #4562, opened 2026-05-04
fix(nodes): deprecate serpapi tool node

RoRoJ:
Add deprecation notice for SerpAPI tool node.


its-imad (approved):
Just a tiny suggestion on this, since Vale really isn't a fan of passive voice :D : "A future version will remove this deprecated node. Use the verified SerpApi Official community node instead." ?

RoRoJ:
> Just a tiny suggestion on this, since Vale really isn't a fan of passive voice :D : "A future version will remove this deprecated node. Use the verified SerpApi Official community node instead." ?

I did think about this, but the wording in the admonition is standard, the same deprecation notice is in all the other deprecated nodes. I think if we change it one place, we should change it everywhere. In this case, the passive voice doesn't bother me.

its-imad:
agree

PR #4637, opened 2026-05-14
docs: replace workflow:updateRedactionSetting with two new scopes

yuliia-pominchuk:
**Summary**

Replaces the deprecated `workflow:updateRedactionSetting` scope with two new granular scopes: `workflow:enableRedaction` and `workflow:disableRedaction`.


phyllis-noester (approved):
nice!

PR #4659, opened 2026-05-18
docs: Document API key scopes (DOC-1957)

its-imad:
## Summary

- Expand the **API Scopes** section on [/api/authentication/](https://docs.n8n.io/api/authentication/#api-scopes) from a 4-line stub to a full reference covering **all 67 scopes across 17 resource groups** (community packages, credentials, data tables, executions, folders, insights, projects, security audits, source control, tags, users, variables, workflows, plus row/column/tag sub-resources).
- Move the section to the bottom of the page so the create → use → configure → delete how-to flow stays uninterrupted.
- Cross-link from step 5 of "Create an API key" to the new reference, and add a `/// note` distinguishing API key scopes from [project role scopes](https://docs.n8n.io/user-management/rbac/custom-roles/).

Closes [DOC-1957](https://linear.app/n8n/issue/DOC-1957).

## Source of the scope list

Captured from the API key creation modal in n8n cloud 2.20.7. Scope-to-endpoint mapping derived from `docs/api/v1/openapi.yml` (operation IDs, tags, and endpoint summaries), supplemented with kapa.ai queries for endpoints lacking inline descriptions.

Re-checking the spec confirmed several descriptions directly, so the remaining IAM ask is narrower than the original list.

## Asking IAM for review

Three things I couldn't resolve from the spec alone:

- **`user:enforceMfa`** — written as "Enforce two-factor authentication enrolment on a user." **No public endpoint exists for this in `docs/api/v1/openapi.yml`**, so the behaviour is inferred from the scope name. Does this scope grant per-user MFA enforcement, or instance-wide enforcement toggling?
- **`workflow:activate` — terminology mismatch.** The endpoint `/workflows/{id}/activate` is titled "Publish a workflow" with the note *"In n8n v1, this action was termed activating a workflow."* The UI scope is still named `workflow:activate`, and my description reflects the UI ("Activate or deactivate a workflow"). Should the docs also mention that this is now called "publish/unpublish" in the API?
- **Archive endpoints have no matching scope.** `/workflows/{id}/archive` and `/workflows/{id}/unarchive` exist in the spec, but there's no `workflow:archive` scope in the UI. Are archive operations covered by `workflow:update`, `workflow:delete`, or neither?

Any other corrections welcome.

## Test plan

- [ ] Preview `/api/authentication/#api-scopes` and confirm all 17 H3 sub-sections render with their tables.
- [ ] Confirm the right-hand on-page TOC populates with the 17 resource-group headings.
- [ ] Confirm the cross-links work: step 5 of "Create an API key" → `#api-scopes`, and the note → `/user-management/rbac/custom-roles/`.
- [ ] IAM engineer reviews the three open questions above.

### DIFF
--- a/docs/api/authentication.md
+++ b/docs/api/authentication.md
@@ -191,7 +191,7 @@ The following tables list the scopes available when you create a scoped API key,
 | `user:read` | Retrieve a user. |
 | `user:list` | List users. |
 | `user:changeRole` | Change a user's global (instance-level) role. |
-| `user:enforceMfa` | Enforce two-factor authentication enrolment on a user. |
+| `user:enforceMfa` | Reserved scope. No `/api/v1/` endpoint currently consumes it, so selecting this scope on a Public API key has no public-facing effect. |
 | `user:delete` | Delete a user from the instance. |
 
 ### Variable scopes
@@ -211,9 +211,9 @@ The following tables list the scopes available when you create a scoped API key,
 | `workflow:read` | Retrieve a workflow and its details. |
 | `workflow:list` | List workflows. |
 | `workflow:update` | Update a workflow. |
-| `workflow:delete` | Delete a workflow. |
+| `workflow:delete` | Delete, archive, or unarchive a workflow. |
 | `workflow:move` | Transfer a workflow to another project. |
-| `workflow:activate` | Activate or deactivate a workflow. |
+| `workflow:activate` | Activate or deactivate a workflow. Also referred to as "publish/unpublish" in the public API (`/workflows/{id}/activate` and `/workflows/{id}/deactivate`). |
 
 ### Workflow tags scopes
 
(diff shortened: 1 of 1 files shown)


RoRoJ (approved):

phyllis-noester (approved):

phyllis-noester:
> workflow:activate — terminology mismatch. The endpoint /workflows/{id}/activate is titled "Publish a workflow" with the note "In n8n v1, this action was termed activating a workflow." The UI scope is still named workflow:activate, and my description reflects the UI ("Activate or deactivate a workflow"). Should the docs also mention that this is now called "publish/unpublish" in the API?

I think it makes sense to mention this in the docs, thanks for calling out

> Archive endpoints have no matching scope. /workflows/{id}/archive and /workflows/{id}/unarchive exist in the spec, but there's no workflow:archive scope in the UI. Are archive operations covered by workflow:update, workflow:delete, or neither?

both use `workflow:delete`

> user:enforceMfa — written as "Enforce two-factor authentication enrolment on a user." No public endpoint exists for this in docs/api/v1/openapi.yml, so the behaviour is inferred from the scope name. Does this scope grant per-user MFA enforcement, or instance-wide enforcement toggling?

looks like this is just dead weight / unused; this scope is only used within the rest api

its-imad:
created this [ticket](https://linear.app/n8n/issue/IAM-696/userenforcemfa-api-scope-is-exposed-publicly-but-has-no-public) to follow-up on user:enforceMfa API scope that is exposed publicly but has no public endpoint

PR #4717, opened 2026-05-28
docs: fix auto-handle-on-label workflow

RoRoJ:
This PR makes minor fixes to the auto-handle-on-label workflow, after recent label changes.

- Adds missing "quality:disruptive" label to if statement
- Adds comment to explain what the webhook does
- Uses webhook to add authors of quality:unusable PRs to the warn list via webhook, rather than block them
- Blocks and closes PRs of authors of quality:disruptive PRs


its-imad (approved):
lgtm

PR #4767, opened 2026-06-05
Import / Export docs preview

geemanjs:


docs/import-export/export-packages.md
```diff
@@ -0,0 +1,121 @@
...
+
+/// note | Proposed feature
+The endpoint, options, and defaults on this page come from a draft design and aren't final.
+///
+
+## Export a package
+
+Send a `POST` request to the export endpoint with the workflows you want to include:
+
+```bash
```
konstantintieber, line 19: Needs docs on how to authorize this request

docs/import-export/import-packages.md
```diff
@@ -0,0 +1,130 @@
...
+
+Import a package to add its workflows and dependencies to an instance. Import options control where the content lands, what happens when something already exists, and how strict n8n is about missing dependencies.
+
+/// note | Proposed feature
+The endpoint, options, and defaults on this page come from a draft design and aren't final. Names and defaults may still change.
+///
+
+## Authenticate your request
+
+Import and export run on the [public API](/api/index.md). Authenticate with an API key in the `X-N8N-API-KEY` header, as described in [API authentication](/api/authentication.md). The import runs as the user that owns the key, and respects that user's roles and permissions.
```
konstantintieber: For the OEM use case it'd be good to mention that one can also authenticate using an `Authorization: Bearer <token>`, which would be the case when using token exchange.

docs/import-export/import-packages.md
```diff
@@ -0,0 +1,130 @@
...
+
+| Value | Behaviour |
+|-------|-----------|
+| `new-version` | Updates the existing workflow whose `sourceWorkflowId` matches, and bumps its version. |
+| `fail` | Reverts all parts of the import and returns an error. |
+| `skip` | Skips the conflicting workflow and continues importing the rest of the package. |
+
+## Control workflow activation
+
+Use `workflowActivationPolicy` to control whether imported workflows are active.
```
konstantintieber: Would be good to have docs about who will be the actor of the publish event.
I assume it would be the user that the token belongs to?

docs/import-export/move-a-workflow.md
```diff
@@ -0,0 +1,76 @@
...
+
+This example walks through the most common task: promoting a single workflow from one instance to another, for example from staging to production. The workflow uses a Slack credential that already exists on the target.
+
+/// note | Proposed feature
+The endpoints, request shapes, and responses here come from a draft design and aren't final. See [Request format](#request-format) for what's still open.
+///
+
+## Before you start
+
+Import and export run on the [public API](/api/index.md), so you authenticate the same way you would for any API call: with an API key in the `X-N8N-API-KEY` header. See [API authentication](/api/authentication.md) for how to create a key. You need a key for both the source and the target instance.
```
konstantintieber: Same here. For OEMs they are likely to use `Authorization: Bearer` header and not `X-N8N-API-KEY`

docs/import-export/move-a-workflow.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Step 2: Import the workflow
+
+Send the package to the target instance:
+
+```bash
+curl -X POST https://prod.example.com/api/v1/import-export/import/package \
+	-H "X-N8N-API-KEY: <prod-api-key>" \
+	-F "package=@workflow-package.tar" \
+	-F 'options={ "credentialMatchingMode": "name-and-type", "credentialMissingMode": "must-preexist" }'
```
konstantintieber: I think we should include the activationpolicy explicitly in here.
As this is crucial for the use case of having one instance where editing takes place and another one that is just supposed to run the production executions.

docs/import-export/move-a-workflow.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+Because the credential has a different ID on production, `credentialMatchingMode: "name-and-type"` tells n8n to match it by name and type instead of by ID. `credentialMissingMode: "must-preexist"` tells n8n to fail rather than create a stub if it still can't find a match.
+
+n8n imports the workflow into your personal project on production, links the Slack credential by name and type, and gives the workflow a new ID while recording the original in `sourceWorkflowId`. To control where the workflow lands or how n8n assigns its ID, see [Import a workflow package](/import-export/import-packages.md).
```
konstantintieber: I think it'd be better (closer to the most common use case) if this example would not assume that I want a workflow to end up in the personal project of a user.
I would like to understand how the import can match the same project of the source instance and place it the corresponding project in the target instance

docs/import-export/move-a-workflow.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Common scenarios
+
+Different tasks call for different option combinations. These are the starting points for the most common ones:
+
+| Scenario | Goal | Key options |
+|----------|------|-------------|
+| Promote to an instance that already has the dependencies | Reuse the target's existing credentials and variables | `credentialMatchingMode: "name-and-type"`, `credentialMissingMode: "must-preexist"`, `variableMissingMode: "must-preexist"` |
```
konstantintieber: This table is hard to read on the website, as the "key options" column is too narrow.

RoRoJ:
Hi @geemanjs , the PR looks very comprehensive, but we need to look at where the content should live in the current structure, and also what the status of the feature is (released? beta?). I've pinged you on Slack.

PR #4769, opened 2026-06-05
Support PR: Increasing HTTP Node Payload Size Limit

luendres:
By default, the HTTP Request node in n8n has a payload size limit of 15 MB. If the user attempts to send or receive data exceeding this size, they may encounter errors due to this limit. This PR explains the possibility of increasing the payload size limit, and also alerts on the possible performance impact of doing so.


its-imad (approved):

its-imad (approved):
LGTM

RoRoJ (approved):

PR #4802, opened 2026-06-12
Update: Remove restcountries dependency

digitalchild:
The existing free and unauthenticated endpoint has been removed.  Replaced with new endpoint on our infrastructure that uses the ISO-3166 country codes list.


Joffcom (approved):
URL works and the text is pretty much the same.

PR #4808, opened 2026-06-12
docs: Document S3 storage mode for execution data

ivov:
Ref: https://github.com/n8n-io/n8n/pull/32226


its-imad (changes requested):
Thanks Iván! Looks good and matches the code in #32226. A few points before, tell me if i am wrong : 

- License expiry differs from binary data. Binary data falls back to read-only, but execution data calls process.exit(1) when the mode is s3 and the feature isn't licensed, so n8n won't start at all. Since both share the "Feature availability" box, can we add a line: if the license is missing/expired with N8N_EXECUTION_DATA_STORAGE_MODE=s3, n8n won't start, and the fix is to switch the mode back.

- Pruning differs. The section points to the binary data setup, which requires an S3 lifecycle rule. But execution data is deleted by n8n itself via the normal EXECUTIONS_DATA_* settings. One sentence would help, otherwise people may add an unneeded lifecycle rule that could delete data n8n still uses.

- Style: i.e. isn't allowed. Change "non-destructive, i.e. older executions..." to "non-destructive. Older executions..." or "that is,".

- (Optional) Maybe note that the mode and S3 variables must be set on all instances, including workers, in queue mode.

Points 1 and 3 are the most important for me. Thanks!

its-imad (approved):
lgtm

PR #4872, opened 2026-06-22
docs: clarify which triggers and runs don't count towards execution quotas

alielkhateeb:
## Summary

Makes it explicit which triggers and runs **don't** count towards execution quotas (manual, sub-workflows, error workflows, empty polls, malformed/rejected requests), and cross-links the production-executions explanation to the quota-counting section.

### Changes
- `workflows/executions/index.md`: adds a **Triggers and runs that don't count** subsection after _Execution count by trigger type_.
- `workflows/executions/manual-partial-and-production-executions.md`: notes that production executions count towards the quota and links to _How executions count towards quotas_.

Closes https://linear.app/n8n/issue/LIGO-750

### DIFF
--- a/docs/workflows/executions/index.md
+++ b/docs/workflows/executions/index.md
@@ -16,7 +16,7 @@ There are two execution modes:
 
 ## How executions count towards quotas:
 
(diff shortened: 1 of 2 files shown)


docs/workflows/executions/index.md
MarcL: @alielkhateeb Do we want to remove the sentence here that says the same thing as the section you've added?

alielkhateeb: True, I will remove it from here so it's explicit down there in the list.

alielkhateeb: @MarcL Claude has done it 😄

RoRoJ (approved):

PR #4880, opened 2026-06-23
docs: Document Azure Blob Storage for binary data

ivov:
Ref: https://github.com/n8n-io/n8n/pull/32837


its-imad (approved):
Hey @ivov 👋 The Azure content looks solid

 I checked the blob path, the "container always required" rule, and the available-modes example against the implementation in n8n-io/n8n#32837 and they all match. The cross-page anchor also resolves correctly on the GitBook preview. ✅

One small thing if you want to tidy up while we're here: the S3 hint just above the new section is now stale.

In use-external-storage.md (line 38), the S3 "Enterprise-tier feature" hint still says:

"If your license key expires and you remain on S3 mode, the instance will be able to read from, but not write to, the S3 bucket."

But in base-command.ts the S3 write-mode block now calls process.exit(1) on startup without a valid license —> same behavior as the new Azure block. So S3 no longer degrades to read-only; it refuses to start.

Suggested replacement to match the Azure wording:

"You will need an Enterprise license key for external storage. n8n won't start in s3 binary data mode without a valid license: set N8N_DEFAULT_BINARY_DATA_MODE to another mode or upgrade your license."

Thanks! 🙏

its-imad:
Follow-up: the stale S3 license-expiry hint discussed above has been fixed in #4947 (merged) — the S3 wording now matches the Azure behavior documented here. Tracked as DOC-2064.

PR #4905, opened 2026-06-25
fix(docs): rename Alibaba Cloud docs to Qwen Cloud

DawidMyslak:
## Summary
- Rename Alibaba Cloud Model Studio, Chat Model, and credentials docs to Qwen Cloud to match n8n#32632.
- Update credential setup instructions and related resource links to Qwen Cloud docs.
- Update integration navigation labels in SUMMARY.md and nav.yml.


its-imad (approved):
LGTM 
Rename is correct, links to qwencloud.com all resolve, and doc paths/anchors are preserved so no redirects needed on the docs side (node type IDs alibabaCloud / lmChatAlibabaCloud / alibabaCloudApi are unchanged in core, so URLs stay stable).

One thing to confirm before/after merge: credential setup step 8 says "Set Region to Singapore. Other regions are available only through Alibaba Cloud Model Studio." The credential still ships all five regions (Singapore, US, China, HK, Germany), so just confirming that's intentional guidance for Qwen Cloud accounts rather than a behavior limit. Non-blocking.

Follow-up for the n8n.io marketing pages (rename + 301 redirects) is tracked for the growth team: GRO-371 — https://linear.app/n8n/issue/GRO-371. The "Docs + Examples" buttons there point at the stable docs URLs, so they keep working regardless.

Minor FYI: the nav.yml change is fine but that file is MkDocs-era residue the GitBook build ignores -> no action needed.

DawidMyslak:
Updated the wording to clarify this is guidance specifically for Qwen Cloud accounts:

> For Qwen Cloud accounts, set **Region** to **Singapore**. Select another region only if you're using Alibaba Cloud Model Studio.

This matches the product change in n8n#32632: the credential still exposes the existing regions, but Qwen Cloud usage is currently Singapore-focused while the other regions remain for Alibaba Cloud Model Studio.

PR #4922, opened 2026-06-29

its-imad:


its-imad:
Closing as redundant: the version snippet (stable 2.27.4 / beta 2.28.2) is already on main via #4921. This PR's content is identical, so it's a no-op. When 2.27.5/2.28.3 publish, a future run will update the snippet.

PR #4936, opened 2026-06-30
docs: Documentation about the durable scheduler

sovietspaceship:
## Summary

Write a **Scheduler** environment variables page as an operator-oriented guide to the durable scheduler, and registers it in the nav (`docs/deploy/SUMMARY.md`).

The page covers three things:

- **The switch from the in-memory scheduler to the durable scheduler**: what each is, why the durable one exists (runs survive restarts, each run executes once across instances instead of relying on a leader), and the functional changes to expect (timing precision, clock-aligned timing preserved by default, late runs after downtime).
- **The concepts needed to work with it**: the *schedule* vs *run* vocabulary and the four stages a run moves through — materialization, execution, recovery (reaper), and retention — so each environment variable maps to a concept.
- **The full set of `N8N_SCHEDULER_*` variables** (19 total), grouped by stage, plus `N8N_SCHEDULER_TRIGGER_NODE_MODE` and its `legacy`/`new` deviation for "every N seconds/minutes" schedules.

Also surfaces that Schedule Trigger takeover requires `N8N_USE_WORKFLOW_PUBLICATION_SERVICE=true`, and notes the scheduler is off by default and rolling out gradually.

## Related

- Linear: https://linear.app/n8n/issue/CAT-3625


looorent (approved):

docs/deploy/host-n8n/configure-n8n/basic-configuration/use-environment-variables/scheduler.md
```diff
@@ -0,0 +1,94 @@
...
+
+These environment variables configure the durable scheduler, which runs time-based workflows from a database-backed queue instead of from each instance's memory. For what the durable scheduler changes, how to turn it on, and how it works, see [Durable scheduler](../../durable-scheduler.md).
+
+{% hint style="warning" %}
+**Preview feature**
+
+The durable scheduler is a preview feature behind an environment flag. The environment variables and default behavior can change before the feature reaches general availability.
+{% endhint %}
+
+## Environment variables <a href="#environment-variables" id="environment-variables"></a>
```
RoRoJ:
```suggestion
```

docs/deploy/host-n8n/configure-n8n/durable-scheduler.md
```diff
@@ -0,0 +1,85 @@
...
+The [remaining variables](basic-configuration/use-environment-variables/scheduler.md) only take effect once the scheduler is on. The defaults suit most instances, so change them only to tune timing precision, storage, or load across instances. All durations are in seconds unless stated otherwise.
+
+## How the durable scheduler works <a href="#how-it-works" id="how-it-works"></a>
+
+These terms make the environment variables easier to reason about:
+
+- **Schedule**: a recurring rule, such as a Schedule Trigger node's "every 15 minutes" setting. The scheduler stores each schedule in the database.
+- **Run**: a single firing of a schedule at a specific time. The scheduler records upcoming runs ahead of time as individual rows.
+
+The scheduler moves each run through four stages, and each stage has its own environment variables:
```
RoRoJ:
```suggestion
The scheduler moves each run through four stages, and each stage has its own [environment variables](basic-configuration/use-environment-variables/scheduler.md):
```

PR #4945, opened 2026-07-01
docs: remove self-hosted-only note from OpenTelemetry tracing page

garfik:
## Summary

OpenTelemetry workflow tracing is no longer self-hosted-only (after adding ability to change settings via UI), so the "Feature availability" hint on this page is outdated.

## Test plan
- N/A (docs-only change, single hint block removed)


its-imad:
duplicate of https://github.com/n8n-io/n8n-docs/pull/4938

PR #4954, opened 2026-07-01
fix(mcp): change page in Build to be an overview

RoRoJ:
The "Connect to n8n MCP" page content was in two separate pages, one in Build and one in Connect, with edits in one not reflecting in the other.

This changes the page in "Build" to be a basic overview of what it means to build with MCP, leaving the pages in Connect to give the full detail.


its-imad (approved):
LGTM
Two optional nits (non-blocking):

In What is MCP?, the "custom agent - connects" should use an em-dash (—) to match the rest of the sentence
MCP is defined twice (intro + What is MCP?) —> could trim the intro to avoid repetition
Approving. 🚀

PR #4984, opened 2026-07-06
Update feature availability note for workflow sharing

Denisf88:
Workflow sharing is available on all cloud plans including starter


its-imad (approved):
lgtm

PR #4986, opened 2026-07-06
ci: add internal link checker (DOC-2070)

its-imad:
## What

Adds a CI check for **broken internal `.md` links**, which the external checker (lychee) can't see. In `lychee.toml`, `base = "https://docs.n8n.io"` rewrites relative links to `docs.n8n.io` URLs, and the `exclude` list then drops everything under `docs.n8n.io` — so internal link rot ships undetected.

Resolves DOC-2070.

## How it works

- **`.github/scripts/check_internal_links.py`** resolves each relative link as a local file and applies the GitBook linking rules:
  - `missing-target` — a `.md` link whose target file doesn't exist
  - `no-md-extension` — trailing slash / no `.md` (GitBook falls back to the GitHub source and 404s)
  - `absolute-path` — leading-slash links (render as GitHub-source 404s even when the file exists)
  - `cross-space-relative` — a `../` link resolving into another GitBook space (must use an `app.gitbook.com` URL)
  - Ignores external URLs, `mailto:`, anchors, code blocks, and GitBook-native content-ref/button blocks
- **`.github/workflows/internal-links.yml`** runs it on the Markdown files a PR **changes**, so it fails on newly introduced broken links without blocking on the pre-existing backlog.

## Validation

- Verified against the live site: leading-slash links (e.g. the crypto node's "here" link) render as `github.com/.../blob/main/…` 404s; GitBook `button`/`content-ref` trailing-slash links resolve fine and are correctly excluded (zero false positives).
- Acceptance test: a broken `[...](missing.md)` link fails (exit 1); a valid link passes.
- Running the checker on `main` surfaces **37 pre-existing broken links**, tracked separately in **DOC-2082** (not fixed here to keep this PR to the tooling).

## Notes

- No repo README added — the check follows the existing convention (comments in the workflow YAML), matching lychee/Vale.
- Related: DOC-2073 (trailing-slash migration links), DOC-2082 (the 37-link backlog).

### DIFF
--- a/.github/scripts/check_internal_links.py
+++ b/.github/scripts/check_internal_links.py
@@ -197,12 +197,16 @@ def classify_cross_space(target: str):
     base = DOCS_ROOT / folder
     if page_path == "":
         return None  # link to space root always resolves
+    broken = ("cross-space-broken", f"cross-space target not found in space '{folder}': {target}")
+    # A valid GitBook page path is relative to the space root and never contains
+    # `.`/`..` segments. Reject them so the filesystem check below can't resolve
+    # `..` at the OS level and escape the mapped space folder (false negative).
+    segments = page_path.split("/")
+    if any(seg in ("", ".", "..") for seg in segments):
+        return broken
     if (base / (page_path + ".md")).is_file() or (base / page_path).is_dir():
         return None
-    return (
-        "cross-space-broken",
-        f"cross-space target not found in space '{folder}': {target}",
-    )
+    return broken
 
 
 def classify(md_file: Path, line: int, target: str):
--- a/.github/scripts/check_internal_links.py
+++ b/.github/scripts/check_internal_links.py
@@ -41,6 +41,35 @@ EXCLUDE_FILES = {
     "docs/contribute/style-guide-for-n8n-docs.md",
 }
 
+# The style guide holds the canonical `space folder -> space ID` table used for
+# cross-space (app.gitbook.com/s/<id>/...) links. Parsed at runtime so there's a
+# single source of truth the team already maintains.
+SPACE_ID_TABLE_FILE = REPO_ROOT / "docs" / "contribute" / "style-guide-for-n8n-docs.md"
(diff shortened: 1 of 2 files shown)


its-imad:
Thanks @cubic-dev-ai — valid catch. Fixed in 1dd78d6: `space_of()` now requires 3+ path parts (`docs/<space>/<file>`), so a file directly under `docs/` returns `None` instead of its filename. It was latent (no files currently live directly under `docs/`) but the fix matches the docstring and prevents false `cross-space-relative` findings. Re-ran the checker: totals unchanged, still zero cross-space false positives.

HumanistSerif (approved):

its-imad:
@kartikay-n8n good question on cross-space checks — I added it in 479a6d9.

The checker now handles both cross-space cases:
1. **Relative `../` links that cross a space boundary** — flagged (they don't resolve in GitBook; you must use an `app.gitbook.com` URL).
2. **`app.gitbook.com/s/<id>/...` links** — validated against the space-ID table in the style guide (`docs/contribute/style-guide-for-n8n-docs.md`). It maps the space ID to its docs folder and checks the target page exists. Single source of truth — no hardcoded IDs; if the team adds a space to that table, the checker picks it up.

Space IDs not in the table (e.g. the reusable-content utility space, and `{% include %}` directives, which aren't links) are skipped and reported for transparency.

This surfaced **9 pre-existing broken cross-space links** on `main` (doubled path segments from the migration, links to pages that don't exist) — tracked in DOC-2082, not fixed here to keep this PR to the tooling.

its-imad:
Thanks @cubic-dev-ai — valid, fixed in ab84bde. A GitBook cross-space page path is relative to the space root and never contains `.`/`..`, so I now reject those segments before the filesystem check. This closes the `..` OS-level escape (a malformed `.../s/<id>/../other-space/page` no longer resolves into another space) and correctly flags it as broken. Verified: no regression (still 9 findings) and the escape case is now flagged.

PR #4450, opened 2026-04-10
docs: remove deprecated N8N_RUNNERS_ENABLED variable

mirtafon:
This PR removes all references to the `N8N_RUNNERS_ENABLED` environment variable across the documentation. 

Since this variable is now deprecated and no longer supported in v2.0+, removing it from the Docker and Task Runner installation guides will prevent user confusion and keep the documentation clean and up-to-date.

All related empty lines and table formatting have been carefully adjusted to maintain the structure. 

---
## Summary by cubic
Removed all references to the deprecated `N8N_RUNNERS_ENABLED` env var across the docs and examples. Updated v2.0 migration notes and Docker/Task Runner guides to reflect that task runners are enabled by default.

- **Migration**
  - No action needed on v2.0+. Remove `N8N_RUNNERS_ENABLED` from any Docker or Compose configs if still present.

<sup>Written for commit d0000c881b0b00e1820a40ee364c74d83164bbf1. Summary will update on new commits.</sup>

PR #4995, opened 2026-07-08
fix(docs): update info for deprecated task runners env var

RoRoJ:
## Summary

Documents the deprecation of the `N8N_RUNNERS_ENABLED` environment variable, which is deprecated from version 2.0 onwards but still supported in version 1.x. The env var mentions are kept in place, with deprecation notes added alongside them.

Replaces PR https://github.com/n8n-io/n8n-docs/pull/4450

## Changes

- **`docs/deploy/host-n8n/configure-n8n/set-up-task-runners.md`**
  - Added an `info` hint in the "Configuring n8n container in external mode" section explaining the deprecation.
  - Updated the `N8N_RUNNERS_ENABLED=true` table row description to note it's deprecated from 2.0 and still supported in 1.x.
- **`docs/deploy/host-n8n/install-options/install-with-docker.md`**
  - Added the same `info` hint after the "Starting n8n" command explanation, covering both `docker run` examples.

## Notes

- Existing `N8N_RUNNERS_ENABLED=true` usages are preserved (the Docker examples reference 1.x images, where the variable is still required).
- Hint wording is kept consistent across both pages.


its-imad (approved):
LGTM

PR #5081, opened 2026-07-23
docs: Document AWS system credentials (federated authentication)

Minilfat:
## Summary

Documents the outcome of the [Implement AWS Federated Credentials](https://linear.app/n8n/project/implement-aws-federated-credentials-a61d1af5ae94) project on the AWS credentials page.

Changes to `docs/integrations/builtin/credentials/aws.md`:

- **New admin section "Using AWS system credentials (federated authentication)"**: what system credentials are, the full source chain with detection variables and precedence order (environment variables → EKS IRSA → EKS Pod Identity → ECS task role → EC2 instance profile), per-platform setup pointers, the security warning for `N8N_AWS_SYSTEM_CREDENTIALS_ACCESS_ENABLED`, proxy behavior, and a troubleshooting list.
- **IRSA added to the supported source list** in the AWS (Assume Role) section (previously undocumented, shipped in n8n-io/n8n#22316).
- **AWS China and GovCloud support**: regional STS endpoints, available from n8n 2.29.0 (n8n-io/n8n#32825).
- **Documents two previously undocumented transitional environment variables**:
  - `N8N_AWS_SYSTEM_CREDENTIALS_SDK_SOURCES` (n8n 2.29.0, n8n-io/n8n#32937)
  - `N8N_AWS_LEGACY_SIGNER` (n8n 2.30.0, n8n-io/n8n#33304)
- Notes that credentials are cached and auto-refreshed via the official AWS SDK.
- Short intro under the page title distinguishing the two credential types.

## Related

- Linear project: https://linear.app/n8n/project/implement-aws-federated-credentials-a61d1af5ae94
- Implementation PRs: n8n-io/n8n#32825, n8n-io/n8n#32937, n8n-io/n8n#32990, n8n-io/n8n#33304

---
## Summary by cubic
Documents AWS system credentials (federated authentication) for self‑hosted n8n and how to use them to assume roles securely. Aligns with the Linear “Implement AWS Federated Credentials” project to reduce static key usage on EKS/ECS/EC2.

- **New Features**
  - Added admin section “Using AWS system credentials (federated authentication)”: how to enable with `N8N_AWS_SYSTEM_CREDENTIALS_ACCESS_ENABLED`, discovery order (env → EKS IRSA → EKS Pod Identity → ECS/Fargate → EC2), per‑platform setup, proxy behavior, and troubleshooting.
  - Clarified that system credentials only seed the STS AssumeRole call. A Role ARN is always required (including self‑assume). Added a trust‑policy example where the `Principal` is the IRSA/Pod Identity/task/instance role, plus an `AccessDenied` troubleshooting note.
  - Updated AWS (Assume Role) docs: the AWS SDK auto‑refreshes temporary credentials and contrasts with the static session token on AWS (IAM). Documented regional STS endpoints for AWS China/GovCloud (from n8n 2.29.0), listed IRSA and EKS Pod Identity in supported sources, and noted that `~/.aws` profiles (`AWS_PROFILE`) aren’t supported.
  - Expanded supported nodes for AWS (Assume Role): added AWS Lambda, AWS SNS, and AWS SNS Trigger.
  - Added a short intro distinguishing AWS (IAM) vs AWS (Assume Role) and documented transitional env vars: `N8N_AWS_SYSTEM_CREDENTIALS_SDK_SOURCES` (2.29.0) and `N8N_AWS_LEGACY_SIGNER` (2.30.0).

- **Bug Fixes**
  - Standardized source naming by using lowercase “instance profile” in the EC2 entry for consistency.

<sup>Written for commit 18ec37873aa6604cd0c148c30c71cff166a9f647. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-docs/pull/5081?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/integrations/builtin/credentials/aws.md
```diff
@@ -203,3 +213,91 @@ If system credentials aren't available, provide these manually:
...
+
+This section is for administrators of self-hosted n8n instances.
+
+{% hint style="info" %}
+System credentials access is for self-hosted n8n only. It's not available on n8n Cloud.
+{% endhint %}
+
+When n8n runs on AWS infrastructure, the infrastructure itself can provide an AWS identity. There are no access keys to create, store, or rotate. This is the standard way to authenticate workloads on EKS, ECS, and EC2, and security teams often require it ("zero static keys").
+
+n8n reads these **system credentials** with the official AWS SDK credential providers, and uses them as the starting credentials for the **AWS (Assume Role)** credential when **Use System Credentials** is turned on. This works for all nodes that accept the AWS (Assume Role) credential, including the AWS Bedrock AI nodes.
```
gthanasis, line 231: Verified this section against the implementation and it's accurate throughout. Two gaps worth closing before merge, both about the same thing: a Role ARN is always mandatory.

There's no path where the server's identity authenticates a node directly. `assertValidRoleCredentials` in `packages/nodes-base/credentials/common/aws/utils.ts` throws `Role ARN is required when assuming a role.` unconditionally, and `getSystemCredentials()` is only ever reached as the master credential for an STS `AssumeRole` call.

So an EKS user whose IRSA role already carries the permissions they need must still supply a Role ARN, pointing either at a second role or at the IRSA role itself. Self-assume requires adding the role's own ARN to its trust policy, which is the step people most often miss.

```suggestion
n8n reads these **system credentials** with the official AWS SDK credential providers, and uses them as the starting credentials for the **AWS (Assume Role)** credential when **Use System Credentials** is turned on. This works for all nodes that accept the AWS (Assume Role) credential, including the AWS Bedrock AI nodes.

{% hint style="info" %}
System credentials are always the starting point for a role assumption, never the identity n8n calls AWS with directly. You must still enter a **Role ARN**, and that role must trust the identity n8n runs as.

If the role n8n already runs as has the permissions your workflows need, enter that same role's ARN and allow it to assume itself. Refer to [Trust policy for system credentials](#trust-policy-for-system-credentials).
{% endhint %}
```

docs/integrations/builtin/credentials/aws.md
```diff
@@ -203,3 +213,91 @@ If system credentials aren't available, provide these manually:
...
+
+[EKS Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) is the newer alternative to IRSA. Install the Pod Identity Agent add-on and create a pod identity association between the n8n service account and an IAM role. No OIDC provider setup needed.
+
+#### Amazon ECS and Fargate
+
+Assign a [task role](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html) to the n8n task definition.
+
+#### Amazon EC2
+
+Attach an [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) to the instance running n8n. n8n uses IMDSv2, so instances with IMDSv1 disabled work.
```
gthanasis, line 286: Follow-on from the Role ARN comment: the page has no trust policy example for the federated path, and the existing one under **Setup Steps** uses `"AWS": "arn:aws:iam::SOURCE-ACCOUNT:root"`. That's right for manual STS keys but wrong here, where the `Principal` must be the IRSA, Pod Identity, task, or instance role.

An `AccessDenied` on `sts:AssumeRole` is the most likely failure right after an admin enables the flag, so it's worth a short subsection. Suggesting it here, after the four platform subsections.

````suggestion
Attach an [instance profile](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html) to the instance running n8n. n8n uses IMDSv2, so instances with IMDSv1 disabled work.

### Trust policy for system credentials

The role you enter as **Role ARN** must trust the identity n8n runs as: the IRSA role, the Pod Identity role, the ECS task role, or the EC2 instance role. Use that role's ARN as the `Principal`, not the account root.

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::123456789012:role/n8n-service-account-role"
      },
      "Action": "sts:AssumeRole",
      "Condition": {
        "StringEquals": {
          "sts:ExternalId": "your-unique-external-id"
        }
      }
    }
  ]
}
```

To find the ARN to use as the `Principal`, run `aws sts get-caller-identity` from inside the n8n container or instance.

If the role n8n runs as already has the permissions your workflows need, use that same role as the **Role ARN** and add its own ARN to its trust policy as shown above. AWS requires this even when a role assumes itself.

The role n8n runs as also needs `sts:AssumeRole` permission on the target role. Grant it on the specific roles workflows should use rather than on `*`.
````

File: docs/integrations/builtin/credentials/aws.md [hunk 3]
```diff
@@ -203,3 +213,91 @@ If system credentials aren't available, provide these manually:
...
+| `N8N_AWS_LEGACY_SIGNER` | `false` | Transitional, available from n8n 2.30.0. Set to `true` to sign AWS requests with the legacy `aws4` signer instead of AWS's Signature V4 implementation. Only use as a rollback if AWS requests fail after an upgrade. n8n removes this flag in a future release. |
+
+### Troubleshooting
+
+* **"Access to AWS system credentials disabled, contact your administrator."**: The **Use System Credentials** option is on, but `N8N_AWS_SYSTEM_CREDENTIALS_ACCESS_ENABLED` isn't set to `true` on the server.
+* **The wrong identity is used**: Check the source order above. Environment variables beat IRSA, which beats Pod Identity, ECS, and EC2. A stray `AWS_ACCESS_KEY_ID` in the container is the most common cause.
+* **Credential test fails on EKS**: Verify the service account annotation (IRSA) or the pod identity association, and confirm the pod received the injected environment variables (`kubectl exec <pod-name> -- env | grep AWS`).
```
gthanasis, line 342: One more troubleshooting entry to pair with the trust policy suggestion above, covering the failure an admin is most likely to hit first.

```suggestion
* **Credential test fails on EKS**: Verify the service account annotation (IRSA) or the pod identity association, and confirm the pod received the injected environment variables (`kubectl exec <pod-name> -- env | grep AWS`).
* **`AccessDenied` when calling `sts:AssumeRole`**: n8n found system credentials, but the target role rejected them. Check that the role's trust policy names the identity n8n runs as, and that the **External ID** matches the policy's condition. Refer to [Trust policy for system credentials](#trust-policy-for-system-credentials).
```

gthanasis (commented):
Thanks for writing this up. I led the [Implement AWS Federated Credentials](https://linear.app/n8n/project/implement-aws-federated-credentials-a61d1af5ae94) project behind this feature, so I checked the whole page against the shipped code.

**Everything technical is correct.** I verified each claim against the implementation:

* All three env var names and defaults match `packages/@n8n/config/src/configs/security.config.ts` and `utils.ts`.
* The source order and every detection variable match the chain in `system-credentials-utils.ts`.
* All three version numbers are right. I checked the commits against release tags: the China and GovCloud fix and the SDK sources flag are both in `n8n@2.29.0`, and the signer flag is in `n8n@2.30.0`.
* The proxy behavior, the fresh-credentials-per-request claim, and the note that CLI profiles are deliberately unsupported all match the code and the security review decisions.

I left three inline suggestions. Two of them are the ones I'd want in before merge, and they're really one issue: **a Role ARN is always mandatory**, which the page doesn't say. See the inline comments for detail.

Two smaller things fall outside the diff, so I couldn't attach suggestions to them:

**1. The supported-nodes list for AWS (Assume Role) is missing three nodes** (around line 96)

`AwsLambda`, `AwsSns`, and `AwsSnsTrigger` all use `awsNodeCredentials`, so they accept `awsAssumeRole`. They're already in the AWS (IAM) list above, just absent from this one. Pre-existing, but it's the list readers will use to decide whether federated auth covers their workflow:

```
* [AWS Lambda](../app-nodes/n8n-nodes-base.awslambda.md)
* [AWS SNS](../app-nodes/n8n-nodes-base.awssns.md)
* [AWS SNS Trigger](../trigger-nodes/n8n-nodes-base.awssnstrigger.md)
```

**2. The Setup Steps trust policy needs a pointer** (line 184, `1. Create the IAM Role in the target AWS account.`)

Its `Principal` is the account root, which is correct for Option 2 but misleading now that the page recommends Option 1. Suggest appending:

> This trust policy is for **Option 2: Manual STS Credentials**. If you use **Option 1: Use system credentials**, the `Principal` is the role n8n itself runs as. Refer to [Trust policy for system credentials](#trust-policy-for-system-credentials).

Happy to push these directly to the branch instead if that's easier for you.

docs/integrations/builtin/credentials/aws.md
```diff
@@ -135,14 +144,15 @@ You have two options for providing credentials to make the STS AssumeRole call:
...
 ##### Option 1: Use system credentials (recommended for server deployments) <a href="#option-1-use-system-credentials-recommended-for-server-deployments" id="option-1-use-system-credentials-recommended-for-server-deployments"></a>
 
-Enable this option if your n8n server has AWS credentials configured through:
+Turn on **Use System Credentials** if the environment your n8n server runs in already has an AWS identity. n8n then discovers credentials for the `AssumeRole` call automatically, without storing static keys in n8n. n8n tries these sources in order and uses the first one that returns credentials:
 
-* Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`)
-* EC2 instance profile
-* ECS task role
-* EKS pod identity
+1. Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SESSION_TOKEN`)
```
RoRoJ, line 152: Between this list and the "Sources" in the table under "Supported credential sources", it would be preferable to use identical naming if possible, unless there is a legitimate difference in naming between the two contexts?

docs/integrations/builtin/credentials/aws.md
```diff
@@ -135,14 +144,15 @@ You have two options for providing credentials to make the STS AssumeRole call:
...
+2. EKS IAM Roles for Service Accounts (IRSA)
+3. EKS Pod Identity
+4. ECS or Fargate task role
+5. EC2 instance profile
```
RoRoJ, line 156:
```suggestion
5. EC2 Instance profile
```

docs/integrations/builtin/credentials/aws.md
```diff
@@ -203,3 +213,91 @@ If system credentials aren't available, provide these manually:
...
+| `N8N_AWS_SYSTEM_CREDENTIALS_SDK_SOURCES` | `all` | Transitional, available from n8n 2.29.0. Controls which system-credential sources resolve through the AWS SDK rather than n8n's legacy resolver. Accepts `all`, `none`, or a comma-separated subset of `environment`, `roleForServiceAccount`, `podIdentity`, `containerMetadata`, `instanceMetadata`. Only change this if AWS authentication broke after an upgrade. n8n removes this flag in a future release. |
+| `N8N_AWS_LEGACY_SIGNER` | `false` | Transitional, available from n8n 2.30.0. Set to `true` to sign AWS requests with the legacy `aws4` signer instead of AWS's Signature V4 implementation. Only use as a rollback if AWS requests fail after an upgrade. n8n removes this flag in a future release. |
+
+### Troubleshooting
+
+* **"Access to AWS system credentials disabled, contact your administrator."**: The **Use System Credentials** option is on, but `N8N_AWS_SYSTEM_CREDENTIALS_ACCESS_ENABLED` isn't set to `true` on the server.
+* **The wrong identity is used**: Check the source order above. Environment variables beat IRSA, which beats Pod Identity, ECS, and EC2. A stray `AWS_ACCESS_KEY_ID` in the container is the most common cause.
```
RoRoJ:
```suggestion
* **The wrong identity is used**: Environment variables beat IRSA, which beats Pod Identity, ECS, and EC2. A stray `AWS_ACCESS_KEY_ID` in the container is the most common cause. See [Supported credential sources](#supported-credential-sources)."
```

docs/integrations/builtin/credentials/aws.md
```diff
@@ -118,13 +125,15 @@ AWS Role Assumption allows you to securely access AWS resources by temporarily a
 * **Principle of least privilege:** Grant only the permissions needed for specific tasks
 * **Audit trail:** Better tracking of who accessed what resources
 
+n8n uses the official AWS SDK to make the STS `AssumeRole` call, and requests fresh temporary credentials when it signs node requests, so you don't need to manage their expiry. This differs from the **Temporary security credential** option on the AWS (IAM) credential, where you enter a static session token that n8n can't renew: when that token expires, you must enter a new one.
```
RoRoJ:
```suggestion
n8n uses the official AWS SDK to make the STS `AssumeRole` call. It requests fresh temporary credentials when it signs node requests, so you don't need to manage their expiry. This differs from the **Temporary security credential** option on the AWS (IAM) credential, where you enter a static session token that n8n can't renew: when that token expires, you must enter a new one.
```

docs/integrations/builtin/credentials/aws.md
```diff
@@ -203,3 +213,91 @@ If system credentials aren't available, provide these manually:
...
+* n8n resolves system credentials when it needs them, so rotated or short-lived source credentials (for example, the IRSA token that Kubernetes rotates) are picked up automatically.
+* Reading AWS CLI profiles from `~/.aws` (`AWS_PROFILE`) is intentionally not supported.
+
+#### Amazon EKS with IRSA
+
+[IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) associates an IAM role with the Kubernetes service account that runs n8n. Annotate the service account with the role ARN (`eks.amazonaws.com/role-arn`). EKS then injects `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` into the pod, and n8n exchanges the token with the regional STS endpoint.
+
+#### Amazon EKS with Pod Identity
+
+[EKS Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) is the newer alternative to IRSA. Install the Pod Identity Agent add-on and create a pod identity association between the n8n service account and an IAM role. No OIDC provider setup needed.
```
RoRoJ:
```suggestion
[EKS Pod Identity](https://docs.aws.amazon.com/eks/latest/userguide/pod-identities.html) is the newer alternative to IRSA. Install the Pod Identity Agent add-on and create a pod identity association between the n8n service account and an IAM role. No OIDC provider setup is needed.
```

docs/integrations/builtin/credentials/aws.md
gthanasis: Applied in 2f6f69b21.

docs/integrations/builtin/credentials/aws.md
gthanasis: Applied in 2f6f69b21 as a new `Trust policy for system credentials` section, plus a pointer on the existing Setup Steps policy noting it covers Option 2.

docs/integrations/builtin/credentials/aws.md
gthanasis: Applied in 2f6f69b21.

docs/integrations/builtin/credentials/aws.md
gthanasis: Reverted this to lowercase in 32db8bf78, since cubic flagged the capital as inconsistent with the sources table, the trust policy prose, and item 4 in this same list.

docs/integrations/builtin/credentials/aws.md
```diff
@@ -135,14 +147,15 @@ You have two options for providing credentials to make the STS AssumeRole call:
...
 
-* Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`)
-* EC2 instance profile
-* ECS task role
-* EKS pod identity
+1. Environment variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SESSION_TOKEN`)
+2. EKS IAM Roles for Service Accounts (IRSA)
+3. EKS Pod Identity
+4. ECS or Fargate task role
+5. EC2 Instance profile
```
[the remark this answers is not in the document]
gthanasis: Applied in 32db8bf78.

docs/integrations/builtin/credentials/aws.md
gthanasis: Done, applied as suggested.

docs/integrations/builtin/credentials/aws.md
gthanasis: Done, applied as suggested.

docs/integrations/builtin/credentials/aws.md
gthanasis: Applied, thanks. Two small extra changes on top of your suggestion, could you take a look: I dropped a stray trailing quote mark that was in the suggested text, and I used "Refer to" instead of "See" to match the wording used elsewhere on this page.

docs/integrations/builtin/credentials/aws.md
gthanasis: Good catch, I made the two lists match. Could you check I picked the right direction: I changed the table's Source column to match this list, rather than the other way round, since the list has no separate platform column to lean on.

docs/integrations/builtin/credentials/aws.md
RoRoJ: Great, thanks

gthanasis (approved):

PR #5086, opened 2026-07-23
docs: Add page for using environments with the public API

SamWooler:
## Summary

Originally raised as a documentation-first design artifact for [API-69](https://linear.app/n8n/issue/API-69/timebox-1d-write-a-docs-page-for-how-to-use-the-existing-environments), written *before* the `status` and `push` public API endpoints existed, as a spec to agree on ahead of implementation.

Both endpoints have since shipped in [API-146](https://linear.app/n8n/issue/API-146/implement-status-and-push-endpoints-on-the-public-api-for-source). This PR reconciles the page with what actually shipped, tracked by [API-279](https://linear.app/n8n/issue/API-279/reconcile-environments-via-api-docs-with-the-shipped-statuspush).

Adds a new sub-page under **Administer > Use source control and environments**: *Use environments programmatically with the public API*.

## What the page covers

- The three source control endpoints (`status`, `push`, `pull`) and how they fit together.
- The **check status, then act** loop: call `status` to see what's eligible, then build `push`'s `fileNames` from it before committing.
- A **development to read-only production** promotion flow, with a preview-before-pull step on production mirroring the preview-before-push step on development.
- A **scheduled backup flow** (the Circle / Teradata use case from API-36), including the risk of running it with `force: true` unattended.
- **Handling push conflicts**: the `409` response shape and retrying with `force: true`.

Loom Demonstration of Source Control Public API endpoints:
https://www.loom.com/share/fc1017f89bf547369ace4473860f4ef0

## Status

Ready for review. Reconciled against the shipped API-146 implementation and validated by hand against two local Enterprise-licensed instances (development/production) walking through every documented flow.


docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+**1. Preview the changes on development.** Call `status` with `direction=push` to see what n8n would commit:
+
+```curl
+curl --request GET \
+	--location '<DEV-INSTANCE-URL>/api/v1/source-control/status?direction=push' \
+	--header 'X-N8N-API-KEY: <DEV-API-KEY>'
+```
+
+**2. Push from development to Git.** Send a commit message. Omit `fileNames` to push everything eligible, or pass a subset of the files from step 1 to push only those:
```
SamWooler: **Design note vs. internal DTO:** this flow assumes `fileNames` is **optional** on the public `push` endpoint (omit it to push everything eligible). The current internal `PushWorkFolderRequestDto` makes `fileNames` **required** (`z.array(SourceControlledFileSchema)`, no `.optional()`). The public endpoint needs to relax this for the "omit to push all" behavior described here. Matches the community PR intent (falls back to all eligible files if omitted).

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,178 @@
...
+| `/source-control/push` | `POST` | Commits and pushes local changes to the connected Git branch. |
+| `/source-control/pull` | `POST` | Fetches changes from the connected Git branch into the instance. |
+
+`push` and `pull` are the same actions as the **Push** and **Pull** buttons in the source control menu. `status` matches the file list the UI shows in the push and pull modals before you confirm. The API exposes it as its own endpoint so a script can do the same review the UI does inline.
+
+For the full request and response schemas, refer to the [Endpoint reference](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/api-reference).
+
+### Authentication and scopes
+
+Every request needs an API key sent in the `X-N8N-API-KEY` header. On Enterprise plans, you can scope each key to the minimum it needs:
```
SamWooler, line 36: **Not strictly accurate.** In the current code the `feat:apiKeyScopes` license flag is only wired to telemetry, not to key creation, the `/api-keys/scopes` endpoint, or request-time enforcement, which are all role-based. So a community instance owner can likely create scoped keys too. Kept the Enterprise framing because it matches the (possibly stale) precedent in `authentication.md`, and it's low-risk here since source control is Enterprise-only anyway. Flagging so we can confirm whether the gate still applies.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,178 @@
...
+
+| Scope | Grants |
+|-------|--------|
+| `sourceControl:pull` | Pull changes into the instance. |
+| `sourceControl:push` | Push local changes to Git. |
```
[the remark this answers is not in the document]
SamWooler: The is intentional, this PR represents a documentation artifact for us to design these endpoints in the Public API around.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,178 @@
...
+	--location '<YOUR-INSTANCE-URL>/api/v1/source-control/push' \
+	--header 'Content-Type: application/json' \
+	--header 'X-N8N-API-KEY: <YOUR-API-KEY>' \
+	--data '{
+		"commitMessage": "Scheduled backup",
+		"force": true
+	}'
+```
+
+Omitting `fileNames` pushes every eligible file, so the branch always reflects the current state of the instance.
```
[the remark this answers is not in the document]
SamWooler: Resolved in https://github.com/n8n-io/n8n-docs/pull/5086/commits/5cda6b4bff728481e13efae3828b264633f941a3

docs/administer/use-source-control-and-environments/use-environments-via-api.md
garfik: Do we really want to have such functionality as "just push whatever we have"? Of course, with git it's possible to rollback easily, but still. In case of random error from client side we will push everything and this potentially can do some harm. 

Also, I'm thinking if we could start with `required`, it will be easier to make it `optional` later, but not other way around

alielkhateeb: I agree with @garfik's concern, and if it is an absolute must have to "push everything" I'd aim for an explicit `{ all: true }` or something, otherwise if the user sends `{}` did they really mean push everything or did they just forget to add the `fileNames` property?

SamWooler: Happy to keep `fileNames` required as it is the safer default and we can always relax it to later. Updated the doc so every push example (promotion, scheduled backup, and partial push) explicitly passes `fileNames`, built from the `status` response, rather than relying on omission.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+{% hint style="info" %}
+**Using multiple branches**
+
+If your instances use different Git branches, you can't promote directly. Push from development to its branch, merge that branch into the production branch in your Git provider, then pull on production. Refer to [Copy work between environments](move-work-between-environments.md) for the branch patterns.
+{% endhint %}
+
+## Scheduled backup flow
+
+To version your instance in Git on a schedule, fetch every eligible file and push everything on a timer. This is a common GitOps backup pattern. Run this from a cron job, a CI pipeline, or an n8n workflow on a [Schedule Trigger](https://app.gitbook.com/s/BKcbOzIWja8NfqKDcqHc/builtin/core-nodes/n8n-nodes-base.scheduletrigger):
```
geemanjs: I don't think this is a gitops pattern as of such (or not one i'm familiar with anyway)

SamWooler: Agreed, removed in https://github.com/n8n-io/n8n-docs/pull/5086/commits/259d59f25374b8d9c91290940b305f8b8873e471.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+## What gets pushed and pulled
+
+The API commits and applies the same resources as the UI: workflows, credential stubs, variable stubs, data table schemas, projects, folders, and tags. For the full list and the merge behavior for each resource type, refer to [Push and pull](push-and-pull-changes.md#what-gets-committed).
+
+## Related
+
+* [Push and pull](push-and-pull-changes.md): How push and pull behave, and what gets committed.
+* [Copy work between environments](move-work-between-environments.md): Branch patterns for moving work between instances.
+* [Endpoint reference](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/api-reference): The full source control API schemas.
```
alielkhateeb: This is the full documentation and not only source control API schemas isn't it?

We can have both? something like this?

```suggestion
* [Source control](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/source-control): Request and response schemas for the source control endpoints.
* [Endpoint reference](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/api-reference): The complete public REST API schemas.
```

SamWooler: Good suggestion, applied in https://github.com/n8n-io/n8n-docs/pull/5086/commits/ae8a57522bb3cfc35e66ffacc9bec5ab659f041e.

SamWooler: Added the reference to the source control endpoints in https://github.com/n8n-io/n8n-docs/pull/5086/commits/bffadc02d6b755e9eeaaa998d1f5fe9a61b77e0d.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -211,4 +211,4 @@ The API commits and applies the same resources as the UI: workflows, credential
 
 * [Push and pull](push-and-pull-changes.md): How push and pull behave, and what gets committed.
 * [Copy work between environments](move-work-between-environments.md): Branch patterns for moving work between instances.
-* [Endpoint reference](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/api-reference): The full source control API schemas.
+* [Endpoint reference](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/api-reference): The complete public REST API schemas.
```
konstantintieber: I think we generally call it "public API schemas", not REST API

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -45,13 +45,14 @@ Scoping keys this way is how you enforce direction. A production instance that s
 
 ## Check the status, then act
 
-Don't push blind. `push` requires a `fileNames` array and only commits the files you list there, so call `status` first to see what's eligible. Because `status` and `push` share the same file shape, you can pass its response straight back with no reshaping, whether you're sending all or a subset.
+Don't push blind. `push` requires a `fileNames` array and only commits the files you list there, so call `status` first to see what's eligible.
```
konstantintieber, line 52: Should this be "blindly"?

konstantintieber: I wonder if this sentence is a bit too harsh sounding

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,221 @@
...
+# Use environments programmatically with the public API
+
+You can drive n8n environments from scripts and pipelines using the public API, without opening the n8n UI. This lets you automate promotion between instances, for example moving work from a development instance to production on a schedule or as part of a CI/CD pipeline.
+
+This page explains how the source control API endpoints work together. It assumes you've already connected your instances to Git and understand how push and pull behave in the UI. For that background, read [Push and pull](push-and-pull-changes.md) and [Copy work between environments](move-work-between-environments.md).
+
+{% hint style="info" %}
+**Feature availability**
+
+Source control and environments is an Enterprise feature. The n8n API isn't available during the free trial. Refer to [API authentication](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/authentication) to create a key.
```
konstantintieber: Would also be good to mention the n8n version tag from when these endpoints are available.

konstantintieber: I think we can stick to just the Environments name here. Also I would take the requirement as "you need to know how to create an API key" to access public API endpoint as a given, that doesn't need to be reiterated here.
```suggestion
The Environments feature is exclusive to enterprise licenses.
```

SamWooler: Done in https://github.com/n8n-io/n8n-docs/pull/5086/commits/29b7a4e524a3503e6ca03ece78396987fd7f7182.

SamWooler: Done in https://github.com/n8n-io/n8n-docs/pull/5086/commits/29b7a4e524a3503e6ca03ece78396987fd7f7182.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
SamWooler: Vale describes 'blindly' as a weasel word 😅

docs/administer/use-source-control-and-environments/use-environments-via-api.md
SamWooler: Done in https://github.com/n8n-io/n8n-docs/pull/5086/commits/29b7a4e524a3503e6ca03ece78396987fd7f7182.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,231 @@
+---
+description: Use the n8n public API to preview, push, and pull source control changes between environments.
+layout:
+  description:
+    visible: false
+---
+
+# Use environments programmatically with the public API
+
+You can drive n8n environments from scripts and pipelines using the public API, without opening the n8n UI. This lets you automate promotion between instances, for example moving work from a development instance to production on a schedule or as part of a CI/CD pipeline.
```
konstantintieber: Nit: `drive` seems a bit of a weird verb in this sentence.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
```diff
@@ -0,0 +1,231 @@
...
+
+This page explains how the source control API endpoints work together. It assumes you've already connected your instances to Git and understand how push and pull behave in the UI. For that background, read [Push and pull](push-and-pull-changes.md) and [Copy work between environments](move-work-between-environments.md).
+
+{% hint style="info" %}
+**Feature availability**
+
+Available from n8n version `2.39.0`.
```
konstantintieber: Might be worth being more precise saying the `/api/v1/source-control/status` and `/api/v1/source-control/push` are available from that version.
To avoid the misunderstanding that anything else related to environments is only available from 2.39.0.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
SamWooler: Switched to `manage` in https://github.com/n8n-io/n8n-docs/pull/5086/commits/7f99044b5f8511cf0fc68c8da3738f392f2f7f59.

docs/administer/use-source-control-and-environments/use-environments-via-api.md
SamWooler: Changed this in https://github.com/n8n-io/n8n-docs/pull/5086/commits/1b03f5525781d009146499e35fedc54099a014ba, LMK what you think.

konstantintieber (approved):
Woohoo 🥳 
Love the real-life use case walkthroughs

HumanistSerif (approved):

PR #5133, opened 2026-07-30
ci: post GitBook page-level preview links on docs PRs

its-imad:
## What

Adds a GitHub Action that posts (and keeps updated) a single PR comment with a **deep link to each page a PR changed** in its GitBook preview. GitBook's own checks only link the space landing page, so reviewers currently hunt through the nav to find the edited page.

## How it works

- **Trigger:** `on: status`. Fires when a GitBook preview build succeeds (no polling), resolves the open PR for the commit, and upserts one sticky comment.
- **Only real pages:** a changed `.md` is linked only if it's listed in its space's `SUMMARY.md`. Nav (`SUMMARY.md`), `.gitbook/` assets, non-`.md`, and unlisted files are reported in a footnote, not linked.
- **Reusable content:** changes under `/.gitbook/includes/` are resolved to the pages that include them via `REUSABLE_CONTENT_INDEX.md` (matched by name, then the block's heading), capped for high-fan-out blocks (some are used in 150+ pages).
- **Readable:** link text is the page **title** (frontmatter `title`/`nodeTitle` → first heading), grouped by space, long lists collapsed in `<details>`.

## Security

`on: status` runs with a write token, so the job must not execute code from a (possibly forked) PR head. It checks out the **trusted default branch** (script + reusable index) and overlays only the PR's changed file **contents** via the API — data, never code.

## Caveat: activates after merge

`status`-triggered workflows only run from the default branch, so this can't be exercised from its own PR branch — it takes effect once merged to `main`.

## Verification

Tested offline against **PR #5132**'s real fixtures (statuses + changed files): produces the exact working deep link
`https://docs.n8n.io/build/~/revisions/<rev>/work-with-data/reference-data/reference-previous-nodes` with the page title. Plus synthetic cases: README/group page, 16-page-one-space collapse, multi-space, removed file, small- and large-fan-out reusable blocks (incl. the heading-fallback match resolving to 153 pages, capped at 10 + "…143 more"), `SUMMARY.md`-only, and assets. A `workflow_dispatch` (SHA input) is included for a post-merge dry run on `main`.

Closes DOC-2166

### DIFF
--- a/.github/scripts/gitbook_preview_links.py
+++ b/.github/scripts/gitbook_preview_links.py
@@ -148,6 +148,18 @@ def deep_link(space_info: dict, rel: str) -> str:
     return base + rel if rel else base
 
 
+PROD_BASE = "https://docs.n8n.io/"
+
+
+def production_url(filename: str) -> str:
+    """Live docs URL for a page, derived from its path (top folder == URL
+    segment; README -> folder index). Used for pages GitBook doesn't build a
+    preview for (reusable-content fan-out)."""
+    rel = re.sub(r"\.md$", "", filename[len("docs/"):])
+    rel = re.sub(r"(^|/)README$", "", rel)
+    return PROD_BASE + rel.strip("/")
+
+
 def in_summary(space: str, filename: str) -> bool:
     summary = REPO / "docs" / space / "SUMMARY.md"
     try:
@@ -218,8 +230,8 @@ def page_line(space_info: dict, filename: str) -> str:
 
 
 def render(changed, spaces, index) -> str:
-    direct: dict = {}          # space -> [file]
-    reusable: dict = {}        # space -> [(name, [pages], total)]
+    direct: dict = {}          # space -> [file]  (pages with a real preview)
+    reusable_blocks = []       # (name, [affected page paths], total)
     non_pages = []             # files that aren't pages
     unresolved_reusable = []   # include files we couldn't map
 
@@ -235,17 +247,7 @@ def render(changed, spaces, index) -> str:
             if not pages:
                 unresolved_reusable.append(filename)
                 continue
-            # group resolved pages by the space that actually built a preview
-            by_space: dict = {}
-            for p in pages:
-                sp = space_of(p)
-                if sp in spaces and (REPO / p).exists():
-                    by_space.setdefault(sp, []).append(p)
-            if not by_space:
-                unresolved_reusable.append(filename)
-                continue
-            for sp, pgs in by_space.items():
-                reusable.setdefault(sp, []).append((name, pgs, len(pages)))
+            reusable_blocks.append((name, pages, len(pages)))
             continue
 
         if Path(filename).name == "SUMMARY.md" or "/.gitbook/" in filename:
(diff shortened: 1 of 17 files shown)


its-imad:
Confirmed the reusable-content behavior against merged PR #5131 (a reusable-only change to `latest-next-version.md`):

- GitBook posted **only** `GitBook (./docs/reusable-content)` (the block diff) — **no** preview statuses for the `changelog`/`deploy` spaces that embed the block. So affected pages have no revision to deep-link into.
- Final handling (updated in e78f988): a **Reusable content** section that links the block **diff** and lists the embedding pages as **live** docs.n8n.io URLs (blast radius), capped, with a note that GitBook doesn't preview embedding pages.

Verified offline against PR #5131's real fixtures (14 pages across changelog + deploy, shown 10 + "…and 4 more") and PR #5132 (direct page, deep-linked). Thanks for the pointer — it removed the one open unknown.

its-imad:
Thanks — worked through all 8. Pushed a33732472.

**Fixed:**
- **Pagination (files.json):** now `--paginate --slurp | jq 'add'` so PRs with >100 changed files (multiple JSON pages) parse correctly.
- **Empty overlay on fetch failure:** fetch to a temp file, replace only on a clean, non-empty decode — a failed fetch no longer clobbers the trusted base-branch copy. Also overlay `REUSABLE_CONTENT_INDEX.md` when it's in the PR.
- **`|| true` hiding script errors:** dropped it — a real script error now fails the step (visible/rerunnable); a legitimate "nothing to post yet" is empty stdout + exit 0.
- **Markdown injection via title:** page titles (PR-controlled) are now escaped before embedding in the sticky comment.

**Already fixed before this review:** the multi-space reusable total — reusable was refactored to a flat list (each block lists all its pages once with the true total), so no per-space overcount.

**Declined with evidence — custom slugs (both comments):** in this repo, frontmatter `url:` is the *pre-migration* URL and is stale. GitBook publishes by file path. Verified against the live site:
- `changelog/release-notes-2.x` (path) → **200**; `release-notes/release-notes` (from `url:`) → **404**
- `release-notes/1.x` (from `url:`) → **301** to canonical `changelog/release-notes-1.x`
- `build/manage-workflows/n8n-packages` (path) → **200**; `manage-workflows/export-and-import/n8n-packages` (`url:`) → **404**
- `contribute` (path) → **200**; `contribute-to-n8n` (`url:`) → **404**

So path derivation is canonical and using `url:` would introduce 404s. Kept path derivation; added a stale-`url:` page to the fixtures as a regression guard. Test suite now 29 checks.

its-imad:
Round 2 — pushed f85f23b8a.

**Fixed:**
- **Cleared/empty files reverting to base (SUMMARY.md):** the overlay now detects a failed fetch by `gh`'s exit code rather than by file size, so a legitimately emptied file applies as a valid zero-byte overlay. Clearing a page's `SUMMARY.md` entry is now reflected (no stale membership). Verified locally: empty content → 0-byte file applied; failed fetch → base copy kept.
- **Removed reusable index (or SUMMARY):** the overlay loop now also processes `removed` files and deletes them from the checkout, so a removed `REUSABLE_CONTENT_INDEX.md`/`SUMMARY.md` can't be read stale. The script already degrades to an empty index when the file is absent (reusable changes then go unresolved, not stale) — added a test for that.

**Already handled (re-flag):** the title Markdown-injection comment — `md_escape()` is applied at both title embeddings (the direct-page line and the reusable list), so PR-controlled titles are escaped before entering the trusted comment.

Tests now 31 checks; `runner / gitbook-preview-links` green.

its-imad:
Round 3 — pushed fde6eb423.

- **Renamed files (valid):** the overlay now deletes `previous_filename` from the checkout on a `renamed` change (in addition to overlaying the new path), so a moved `SUMMARY.md`, reusable index, or page isn't read from its stale old location. Verified locally (old path deleted, new content applied).
- **Title injection (re-flag):** already handled — `md_escape()` is applied to every PR-controlled title before it enters the comment. To make that unmistakable I renamed the variable to `safe_title` at both embeddings; no behavior change.

`runner / gitbook-preview-links` green; 31 checks.

its-imad:
Round 4:

- **Renamed files:** already handled — the loop deletes `previous_filename` on a `renamed` change (see the `if [ "$st" = "renamed" ]` branch), then overlays the new path. The re-flag anchored to the `removed` branch just above it; no change needed.
- **P3 — tab/newline/backslash in a renamed path:** declining. `@tsv` would escape such characters, but Markdown doc filenames in this repo can't contain tabs or newlines (and don't contain backslashes), so the input can't occur; the only consequence would be one stale file left undeleted. Not worth NUL-delimited parsing here for a best-effort preview comment.

I think this is a good stopping point — all P2s are addressed (real ones fixed, the rest verified as already-handled or false positives), CI is green, and there's a 31-check fixture test guarding the behavior. Ready for human review.

PR #5111, opened 2026-07-27
docs: add Docker Compose installation options and prerequisites

HumanistSerif:
## Summary by cubic
Makes Docker Compose and the one-line installer the default install paths. Adds a full Compose guide (with bundled `n8n-sandbox` and SearXNG), updates cloud guides, deprecates `npm` installs for v3, and pins the one-line command to an explicit version for stability per Linear DOC‑2087.

- **New Features**
  - New “Install using Docker Compose” guide with prerequisites (Compose v2, 4 GB/2 vCPU), WSL notes, mTLS bootstrap, health checks, optional PostgreSQL, a security checklist, and a service diagram; uses GHCR images for the sandbox and exposes only port 5678.
  - New “One-line setup” that scaffolds `./n8n`, starts n8n with bundled `n8n-sandbox` and SearXNG, documents start/stop/upgrade/uninstall commands and flags, checks for Docker Compose v2, includes Windows/WSL notes, and now requires `--version` to pin the installed n8n version.
  - AI Assistant setup now “Pick your setup” with 3 paths: bundled `n8n-sandbox` (one‑line/Compose), self‑hosted `n8n-sandbox` (advanced), or Daytona (recommended for production); SearXNG is default search with an option for Brave; improved troubleshooting and disable steps.
  - DigitalOcean and Hetzner guides add an optional AI Assistant section to include the sandbox stack without exposing ports (only n8n is proxied via Caddy) and add sizing guidance.

- **Bug Fixes**
  - Standardized AI Assistant env vars and examples (use `N8N_INSTANCE_AI_*` and `N8N_INSTANCE_AI_SANDBOX_API_URL`/`N8N_INSTANCE_AI_SANDBOX_API_KEY`), switched image references to `docker.io/n8nio/n8n` to avoid pull rate limits, marked “Install with Docker” as outdated, added `npm` v3 deprecation and “AI Assistant not available on `npm`,” and fixed broken links and minor copy issues.

<sup>Written for commit 8d64ceddd2c248b82947510391a3452eeeb958c7. Summary will update on new commits.</sup>

[...]

PR #5137, opened 2026-07-31
preview-links: show 'building' state for pages whose space isn't ready yet

its-imad:
## What

Follow-up to #5133 (DOC-2166). When a PR changes pages in a space whose GitBook preview build hasn't finished yet, those pages were reported as *"not standalone pages"* — misleading, since they're real pages the build just hadn't produced.

Now the script distinguishes three states for a changed page:
- space build **succeeded** → deep link (unchanged);
- space has a **pending** GitBook status → a `⏳ GitBook is still building the preview for <space>` note (the sticky comment updates when the build lands);
- space has **no** GitBook status and the file isn't in `SUMMARY.md` → genuinely unlisted / non-page.

## Why

Found by dry-running the live workflow on an open PR (#5111, deploy pages, `deploy` still building): 2 real pages (in `SUMMARY.md`) were lumped with 2 genuinely-unlisted added pages under "aren't standalone pages." After this change, the 2 real pages show as *building* and only the 2 not-in-SUMMARY pages are flagged.

## Changes
- `gitbook_spaces()` — set of spaces with any GitBook context (success or pending).
- `render()` — a `pending` bucket + note; only classifies as non-page when the space has no status at all.
- Tests: +2 checks (pending-space detection and rendering). 33 total.

Verified offline against #5111's real fixtures (with the workflow's overlay) and the fixture suite.

### DIFF
--- a/.github/scripts/gitbook_preview_links.py
+++ b/.github/scripts/gitbook_preview_links.py
@@ -133,12 +133,15 @@ def load_spaces(status_json) -> dict:
 
 
 def gitbook_spaces(status_json) -> set:
-    """All spaces that have any GitBook status (success OR pending). Used to tell
-    'this space's preview is still building' apart from 'not a page' — a page
-    whose space is only pending must not be reported as a non-page."""
+    """Spaces whose GitBook build is success or pending (in progress). Used to
+    tell 'this space's preview is still building' apart from 'not a page'. Failed
+    builds are excluded — a page there shouldn't promise an update that never
+    lands (it falls through to the non-page footnote instead)."""
     statuses = status_json.get("statuses", status_json) if isinstance(status_json, dict) else status_json
     out = set()
     for s in statuses:
+        if s.get("state") not in ("success", "pending"):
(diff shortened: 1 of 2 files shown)


its-imad:
Both addressed — pushed 6899bb674.

- **Failed builds treated as "building":** `gitbook_spaces()` now counts only `success`/`pending` states, so a failed GitBook build no longer shows a "still building" note that will never resolve; those pages fall through to the non-page footnote.
- **All-pending produced no comment:** `main()` now computes the pending set before the early-exit, so a run where no space has succeeded yet (e.g. a manual dispatch during the pending window) still renders the "⏳ building" note instead of emitting nothing.

Added tests for both (failed build → not pending; all-pending → building note). 35 checks.

One deliberate non-change: the workflow still triggers only on `success` status events, not `pending`. Firing on every pending event would multiply runs across all PRs for little gain; in practice the first space to succeed triggers a run that already renders the still-pending spaces as "building," and the comment updates as the rest land.

PR #5147, opened 2026-08-03
docs: update reusable content index (DOC-2183)

its-imad:
Fixes DOC-2183

Updates `REUSABLE_CONTENT_INDEX.md` (109 reusable blocks) — ID, name, parent space, include snippet, preview, and pages that use each block.

Auto-generated by the DocFather Reusable Content Index workflow. Linear: https://linear.app/n8n/issue/DOC-2183


its-imad:
Closing as a false positive. The change here (dropping `docs/integrations/builtin/credentials/kibana.md` from the `cred-only-statement` "Used in" list) is not a real content change — that page still contains the `{% include .../reusable/7QbEnpnpOks3Rq0SiMFb/ %}` on `main` (last edited in #4903, 24 Jun). A deterministic `git grep` finds 36 files referencing the block, kibana.md included.

Root cause: the DocFather Reusable Content Index spoke enumerated references via the GitHub code-search API, which is eventually-consistent and paginated without a stable sort, so a real reference can drop out of the result set between runs. Fixing the spoke to enumerate deterministically via the Git Trees API. Closing this PR (and DOC-2183) to avoid churn.

PR #5048, opened 2026-07-14
docs(changelog): rename feature feed to Release notes, feature-first changelog headlines

Johannesishere:
Implements the release-communication naming agreed by PMM on 2026-07-14 (three tiers: **Changelog** = curated narrative · **Release notes** = every feature, one line each · **GitHub releases** = full technical detail).

## What changed

- **Feature feed → Release notes.** `docs/changelog/feature-feed.md` renamed to `docs/changelog/release-notes.md` (new URL `/changelog/release-notes`), nav label updated in `SUMMARY.md`. The page shipped last week, so nothing external points at the old slug — replaced, not redirected.
- **Update/versioning preamble moved in.** The "How to update n8n" reusable include and the semantic-versioning explainer from the archived 2.x release notes now sit at the top of the Release notes page, which takes over as the living per-version document. The archived 2.x/1.x/0.x pages are untouched.
- **Stable anchors per release.** Every release section on the Release notes page gets an explicit `id="n8n<digits>"` anchor (`#n8n230` for 2.30 — same convention the old MkDocs release-notes page used and the n8n editor's updates panel still links with).
- **Changelog headlines are now feature-first.** Version numbers moved out of the 24 entry headlines into the `**Released:**` line, which links to that version's section on the Release notes page, e.g. `**Released:** 2026-07-07 in [n8n 2.30](release-notes.md#n8n230)`.
- **RSS feed renamed** to `feeds/release-notes.xml` (title/link/self-link updated inside). Same reasoning as the slug: days old, no subscriber base to protect.
- **"Check the Release notes for breaking changes"** bullet in the `updating-best-practices` include now points at GitHub releases, where breaking changes are actually documented (it previously pointed at the frozen 2.x archive). All three duplicate include copies updated.

## 🙏 Redirect request for the docs team (GitBook site settings)

> **Corrected 2026-07-15:** this section originally claimed the `/release-notes/*` URLs were 404ing. They are not — the redirects from the space rename are in place and working (verified with curl: `/release-notes` 307s to the 2.x archive, `/release-notes/1-x` and `/0-x` to their archives). Thanks Rowena for the catch.

One redirect retarget, please apply **after this PR merges**:

| Source | Current target | New target |
|---|---|---|
| `/release-notes` | `/changelog/release-notes-2.x` | `/changelog/release-notes` |

[...]

PR #5156, opened 2026-08-04
docs(changelog): Schedule Trigger runs survive restarts and scale across mains (n8n 2.32)

fabian8n:
Adds the Schedule Trigger runs survive restarts and scale across mains changelog entry for n8n 2.32 (REL-73).

🤖 Generated with Claude Code


Johannesishere:
Review notes, verified against n8n-docs `origin/main` (`1c0e86692`) and n8n `origin/master` (`40dac4b90a`, 2.34.0). The prose is good; the format is the pre-#5048 shape, and two facts belong to a different version.

## Blockers

**1. Wrong place, wrong heading level.** The entry is appended at the bottom of `docs/changelog/README.md`, after the 2.1 "Time Saved node" entry, and as `###` — so GitBook renders it as a *subsection of the 2.1 entry*. Entries are newest-first, directly after the `{% include %}` block and its `***`, so 2.32 goes above "App-only authentication for Microsoft nodes" (2.30). cubic flagged this too.

**2. Retired heading + Released format.** `### n8n 2.32 — Schedule Trigger runs…` is the format PR #5048 retired: no version badge in the heading, no em dash, `##` not `###`, and `**Released:**` needs the version link: `**Released:** 2026-07-21 in [n8n 2.32](release-notes.md#n8n232)`. The date and the `#n8n232` anchor both check out.

**3. `N8N_SCHEDULER_MISFIRE_GRACE` isn't in 2.32.** It landed 2026-07-30 in n8n-io/n8n#35075; `git tag --contains 8672d9567c` returns `n8n@2.34.0`, so it shipped in 2.34, not 2.32 (07-21). The closing sentence goes with it: misfire coalescing ("only the most recent missed occurrence fires") is that same PR's behavior. In 2.32, a missed run simply fires late. Cut both here and let them carry a 2.34 entry once `release-notes.md` has a `#n8n234` anchor. (cubic read this as a docs gap; it's really a version-attribution error — the docs page is correct for 2.32.)

**4. Docs link must be a GitBook space URL.** `https://docs.n8n.io/deploy/…/durable-scheduler` renders as an external link. Use `https://app.gitbook.com/s/jm0ZYRpZIPWge2ZSiDYO/host-n8n/configure-n8n/durable-scheduler`.

**5. Preview caveat needs a hint block.** A preview feature is a sanctioned exception only with a `{% hint style="warning" %}`. Right now the caveat is the fourth sentence of the last paragraph, where it reads as a footnote. An availability block is also missing (`**Availability:** Self-hosted only.`).

**6. CI is red on a pre-existing bug this PR inherits.** `runner / internal-links` fails on README lines 160 and 164: the MCP Client links say `n8n-nodes-langchain.mcpclient`, but the file is `n8n-nodes-langchain.mcpClient.md`. Not this PR's text, but the checker runs on changed files, so this can't go green until the casing is fixed. One character in each, worth folding in here.

## Should fix

**7. The dispatch-timing sentence contradicts the page it links to.** The entry says dispatch "fires at the exact scheduled instant… not limited to a polling cadence." `durable-scheduler.md:39` says the opposite: "a run can start up to one interval after its scheduled time. The in-memory scheduler fires closer to the exact instant." The **entry is right**: `packages/@n8n/scheduler/src/core/executor/precision-timer.ts` (shipped in 2.31) claims each run slightly before `runAt` and fires it on a timer, "steady-state precision is milliseconds." So the docs page is stale, and so is the `N8N_SCHEDULER_EXECUTOR_INTERVAL` description in `scheduler.config.ts`, which still calls itself the "worst-case delay." Both need a follow-up, and the entry shouldn't merge while it contradicts the page it links to.

**8. The Grafana dashboard isn't a findable artifact.** "import the n8n Durable Scheduler Health Grafana dashboard" has no link, and nothing by that name exists in n8n or n8n-docs. Either link it or drop it and point at [Visualize metrics with Grafana](https://docs.n8n.io/deploy/host-n8n/keep-n8n-running/visualize-metrics-with-grafana/). `N8N_METRICS_INCLUDE_SCHEDULER_METRICS` itself is real and shipped in 2.31, but it's undocumented in the scheduler env-var page — also worth a follow-up.

**9. Length and one contraction.** ~430 words against the 200–300 target; the env-var list is what the linked page is for. And "before it is due" → "before it's due" (cubic).

## Rewritten entry, to sit at the top of the page after the `{% include %}` line and its `***`

```markdown
***

## Schedule Trigger runs survive restarts and scale across mains

**Released:** 2026-07-21 in [n8n 2.32](release-notes.md#n8n232)

You can now run Schedule Trigger nodes from a database-backed queue instead of from each instance's memory. The durable scheduler records every upcoming run in the database before it's due, so a restart or crash no longer silently drops work. Any run whose time passed while an instance was down fires when the instance recovers rather than being skipped.

In a multi-main setup, every main instance shares the same queue and claims runs from it. Only one instance picks up each run, so scheduling load spreads across your whole fleet instead of depending on a single leader instance. Dispatch stays precise: the scheduler claims each run slightly ahead of time and holds it on a timer, so steady-state timing isn't limited to a polling cadence.

To opt in, set two environment variables and restart your instance:

1. Set `N8N_SCHEDULER_ENABLED=true` to enable the durable scheduler.
2. Set `N8N_USE_WORKFLOW_PUBLICATION_SERVICE=true` so the scheduler takes over Schedule Trigger nodes. Without this second flag, Schedule Trigger nodes keep running on the in-memory scheduler and n8n logs a warning.

Your existing Schedule Trigger nodes switch automatically without republishing. The defaults suit most instances, and you can tune materialization, timing, retention, and load across instances with the [scheduler environment variables](https://app.gitbook.com/s/jm0ZYRpZIPWge2ZSiDYO/host-n8n/configure-n8n/basic-configuration/use-environment-variables/scheduler). To monitor it, set `N8N_METRICS=true` and `N8N_METRICS_INCLUDE_SCHEDULER_METRICS=true` to expose Prometheus metrics for queue depth, scheduling lag, dispatch throughput, retries, and dead-letters.

Learn more in the [durable scheduler documentation](https://app.gitbook.com/s/jm0ZYRpZIPWge2ZSiDYO/host-n8n/configure-n8n/durable-scheduler).

{% hint style="warning" %}
**Preview feature**

The durable scheduler is a preview feature behind an environment flag. The environment variables and default behavior can change before general availability. n8n recommends testing it in a non-production environment first.
{% endhint %}

{% hint style="info" %}
**Availability:** Self-hosted only.
{% endhint %}
```

Two follow-ups this surfaces, both separate PRs: the stale timing claim on the durable scheduler page (plus the `N8N_SCHEDULER_EXECUTOR_INTERVAL` description in `@n8n/config`), and adding `N8N_METRICS_INCLUDE_SCHEDULER_METRICS` and `N8N_SCHEDULER_MISFIRE_GRACE` to the scheduler env-var page.

sovietspaceship (approved):

PR #4002, opened 2025-12-11
docs: add rustfs as object storage support for S3 node

majinghe:
Recently, minio announced the [minio/minio](https://github.com/minio/minio) repo will be under maintenance mode, [RustFS](https://rustfs.com) is as an alternative to replace the minio. RustFS is also an object storage open source project licensed with Apache 2.0, and now has more than 16k star. So add the RustFS as a object storage support for S3 node.

---
## Summary by cubic
Added RustFS to the S3 node and credentials docs as a supported S3-compatible object storage provider.

<sup>Written for commit 8c4715025afc069347143d42d20a9cdfa479cf79. Summary will update automatically on new commits.</sup>

PR #5169, opened 2026-08-06
docs: add RustFS documentation for S3 credentials and node

majinghe:
## Description

As discussed before, [RustFS](https://github.com/rustfs/rustfs) is a S3-compatible object storage and is one of the MinIO alternative since MinIO has archived its open-source project.

I have open a PR https://github.com/n8n-io/n8n-docs/pull/4002 when RustFS is under alpha, but now we have prompted to beta and will release rc within one week. So i create a new PR to add support for RustFS in n8n.

One more thing, [RustFS](https://rustfs.com) has more than 1.5M installations globally, and the Docker Hub pull also exceeds 7M now. I think, the RustFS can be another good choice for n8n user which already has RustFS instance.


RoRoJ:
Hi @majinghe 

Thanks for the follow-up PR, and for the details on RustFS's growth since #4002.

Since your last submission, we've tightened our contribution guidelines around this sort of case, see https://docs.n8n.io/contribute/contribution-guide-for-n8n-docs#what-not-to-submit

> Promotional or commercial content of any kind, including adding a provider or tool to a curated list of compatible options. Such lists in the doc usually cover only the most widely-used options and are not exhaustive directories — we aim to be vendor-neutral and only expand them when there's a clear reason to do so.

This PR falls into that category.

Closing this for now - thanks again for the interest in supporting n8n users on RustFS.

PR #5174, opened 2026-08-06
docs(changelog): Configure instance settings programmatically via REST API (n8n 2.31)

fabian8n:
Adds the Configure instance settings programmatically via REST API changelog entry for n8n 2.31 (REL-76).

🤖 Generated with Claude Code


Johannesishere:
Review notes, verified against n8n-docs `origin/main` (`4fdb7e8dc`) and n8n `origin/master` (`6ab25697e4`, 2.34.0). The API facts are almost all correct and well checked, but one paragraph describes a different endpoint's behaviour, and the format is the pre-#5048 shape again.

## Blockers

**1. Wrong place, wrong heading level.** The entry is appended at the bottom of `docs/changelog/README.md`, after the 2.1 "Time Saved node" entry, and as `###` — so GitBook renders it as a *subsection of the 2.1 entry*. Entries are newest-first. Since #5148 merged, 2.31 (2026-07-14) now sits below the 2.33 SharePoint Excel entry and above the 2.30 "Capture who approved and when" entry.

**2. Retired heading + Released format.** `### n8n 2.31 — Configure instance settings…` is the format PR #5048 retired: no version badge in the heading, no em dash, `##` not `###`, and `**Released:**` needs the version link: `**Released:** 2026-07-14 in [n8n 2.31](release-notes.md#n8n231)`. The date and the `#n8n231` anchor both check out. There's also no `***` divider before the entry.

**3. The redaction paragraph describes a different endpoint.** This is the blocking one:

> Sensitive fields are redacted in GET responses; sending back the redacted placeholder in a PUT keeps the stored value unchanged, while providing a new value updates it.

The security policy has no sensitive fields. `security-policy.yml` is two booleans (`personalSpacePublishing`, `personalSpaceSharing`), a `redactionEnforcement.floor` enum, and three read-only integer usage counts. Nothing is masked. Redacted-placeholder round-tripping is **OIDC** behaviour — `OIDC_CLIENT_SECRET_REDACTED_VALUE` in `public-api/v1/handlers/sso-oidc/sso-oidc.mapper.ts` — and the OIDC endpoints shipped in 2.34, not 2.31. It reads like "execution-data **redaction** enforcement" (a policy setting this endpoint writes) got conflated with response masking.

The genuinely useful fact it displaced, straight from the spec: the read-only usage counts GET returns are *ignored on write*, so a GET response can be sent straight back as a PUT body. That's worth a sentence; the redaction claim needs to go.

**4. Docs link must be a GitBook space URL.** `https://docs.n8n.io/connect/n8n-api/security-policy#retrieve-the-security-policy` renders as an external link. Use `https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/security-policy`.

Two notes on that link: the page exists on the GitBook side (live URL returns 200, a control URL 404s) but hasn't synced into `main`'s tree yet, so it won't show up if you grep the repo. And I couldn't verify the `#retrieve-the-security-policy` anchor — GitBook client-renders, so it isn't in the served HTML. Worth one click-through before merge; I've dropped it from the corrected entry below.

**5. Missing availability hint.** Both handlers are gated on `isLicensed('feat:personalSpacePolicy')`. This isn't available on every plan, so it needs a hint block. The 2.8 entry already sets the precedent: "Personal space policies … are available on Enterprise."

## Should fix

**6. "Rolling out in subsequent releases" is already stale.** All five have shipped: log streaming and SAML in 2.32, OpenTelemetry in 2.33, OIDC and LDAP in 2.34. Future tense was accurate when the memo was written on 07-14, but the page is read today. Name the versions instead.

## Verified and correct

Flagging these so they don't get re-litigated: the endpoint shipped in 2.31.0 (`f54893f0b4`, 2026-07-13, first tag `n8n@2.31.0`); both paths are right; `securitySettings:manage` is a hard requirement (the middleware checks `tokenGrant.apiKeyScopes` directly — no global fallback, despite what the helper is called); scopes are persisted on the key row at creation, so "existing keys do not pick up new scopes" holds; partial bodies do 400 (`required` plus `additionalProperties: false`); 409 when env-managed; and "same validation as the UI" is the spec's own wording. Clean on em dashes in the prose.

## Corrected entry

Remove the appended block at the end of the file and insert this after the 2.33 SharePoint Excel entry, before the 2.30 "Capture who approved and when in human-in-the-loop steps" entry:

```markdown
***

## Configure instance settings programmatically via REST API

**Released:** 2026-07-14 in [n8n 2.31](release-notes.md#n8n231)

You can now read and update instance-level settings through the n8n public REST API, starting with the security policy. Instead of clicking through the settings panel on every instance, you can drive configuration from a CI/CD pipeline, an infrastructure-as-code tool, or any HTTP client. Every write goes through the same validation as the UI, so an API-driven change takes effect exactly as it would if you clicked Save.

`GET /api/v1/settings/security-policy` and `PUT /api/v1/settings/security-policy` read and update personal-space publishing and sharing, plus the execution-data redaction enforcement floor. The PUT takes the full configuration object: partial bodies are rejected with a 400, so you always know exactly what you're writing. The read-only usage counts that GET returns are ignored on write, so you can send a GET response straight back as a PUT body. If the policy is managed through environment variables, the API returns a 409 Conflict and leaves the values untouched, so declarative config is never silently overwritten.

To use these endpoints, generate an API key with the `securitySettings:manage` scope. Scopes are fixed on a key when you create it, so an existing key won't pick up a new scope: issue a new one. Settings endpoints for log streaming and SAML followed in 2.32, OpenTelemetry in 2.33, and OIDC and LDAP in 2.34.

Learn more in the [security policy API documentation](https://app.gitbook.com/s/r7wKI4I1BgdBCuq5Cvcx/n8n-api/security-policy).

{% hint style="info" %}
**Availability:** Enterprise.
{% endhint %}
```

## One process note

This is the fourth changelog PR in a row landing in the retired pre-#5048 shape (bottom of file, `### n8n X.YZ — …` heading, `**Released:**` with no version link). Same note as on #5151: something upstream in the generation workflow is still on the old convention and it's worth fixing at the source rather than in review each time.

Also worth knowing for merge order: this and #5156 both insert into the same region near the top of the page, so whichever lands second will need a rebase. Both are also currently branched off a stale `main` — a rebase clears the red `runner / internal-links` check, which was the `n8n-nodes-langchain.mcpclient` casing bug and is already fixed on `main`.

Johannesishere:
@fabian8n A thorough review comment above. Maybe you can feed it to Claude and AIA to improve the WF?

fabian8n:
Applied, rebuilt on current main as a single commit. Changes: moved between 2.33 and 2.30, `##` heading with the version link on Released, redaction paragraph replaced with the ignored-usage-counts fact from the spec, follow-on versions named (verified via tags: 2.32 log streaming + SAML, 2.33 OTel, 2.34 OIDC + LDAP), GitBook space URL for the docs link, Enterprise hint added.

One deviation: no `***` divider. There are none anywhere on the current changelog page, so I left it out. The `#retrieve-the-security-policy` anchor does resolve on the live page, but I dropped it as suggested.

On the process note: the generation workflow switched to the post-#5048 shape on 08-11; this PR was generated on 08-06, so it predates the fix. Related: the 2.32 Schedule Trigger entry from #5156 is on main in the old shape at the bottom of the file. I'll send a small follow-up for that.

Johannesishere (approved):
All five blockers applied and verified: placement between 2.33 and 2.30, `##` heading with the version link on Released, redaction paragraph replaced with the ignored-usage-counts fact, follow-on versions named, GitBook space URL, Enterprise hint. Agreed on dropping the `***` divider — there are none on the page today, my review asked for something the changelog doesn't do.

PR #5176, opened 2026-08-06
docs: Expand n8n packages

geemanjs:
Update n8n packages documentation to better match the feature


docs/connect/n8n-cli.md
```diff
@@ -176,16 +176,17 @@ n8n-cli workflow transfer <id> --project=<projectId>
...
-n8n-cli package import --file=export.n8np --conflict-policy=fail
-n8n-cli package import --file=export.n8np --project=<project-id> --workflow-id-policy=source --conflict-policy=skip
+n8n-cli package export --project-id=<project-id> --output=project.n8np
+n8n-cli package import --file=export.n8np --workflow-conflict-policy=fail
+n8n-cli package import --file=export.n8np --project-id=<project-id> --workflow-conflict-policy=skip
 ```
 
-See [n8n packages](https://app.gitbook.com/s/rPN1zU5jaYNvwH7RzxqA/manage-workflows/n8n-packages) for a detailed overview of the available flags and what makes up a package.
\ No newline at end of file
+See [n8n packages](https://app.gitbook.com/s/rPN1zU5jaYNvwH7RzxqA/manage-workflows/n8n-packages) for what makes up a package, and [Import a package](https://app.gitbook.com/s/rPN1zU5jaYNvwH7RzxqA/manage-workflows/n8n-packages/import-a-package) for every available flag.
```
ireneea, line 192: This are preview URLs right? should we replace them with the real ones?

docs/build/manage-workflows/n8n-packages/export-a-package.md
```diff
@@ -0,0 +1,200 @@
...
+
+# Export a package
+
+{% hint style="info" %}
+**Feature availability**
+
+n8n packages are in preview. The export options may change in future releases.
+{% endhint %}
+
+Export writes an [n8n package](README.md), a [`.n8np` archive](package-format.md), containing what you asked for, plus references to everything those workflows need on the target instance. Export doesn't modify the source instance.
```
ireneea: [nit] The code formatting for '.n8np' shows a bit strangely inside as a link 

```suggestion
Export writes an [n8n package](README.md), a [.n8np archive](package-format.md), containing what you asked for, plus references to everything those workflows need on the target instance. Export doesn't modify the source instance.
```

docs/build/manage-workflows/n8n-packages/export-a-package.md
```diff
@@ -0,0 +1,200 @@
...
+## Errors
+
+| Status | Cause |
+|--------|-------|
+| `400` | A requested workflow, folder, or project doesn't exist, or you can't access it. Also mixing ID groups, passing no IDs, exceeding 300 IDs in a group, or a blocked export under `--missing-workflow-dependency-policy=fail` |
+| `403` | The API key is missing a scope it needs |
+
+A missing entity and an inaccessible one return the same `400` with the same message, so you can't use export to test whether an ID exists. Export never returns a `404`.
+
+Error bodies carry a `message` only, and it names how many entities were affected rather than which. The CLI prints the same message, and the log streaming event records only the failure reason and the IDs you asked for, so the offending IDs aren't surfaced anywhere.
```
ireneea, line 173: Technically for some errors we do show a subset of the offending IDs 😅

ireneea: Actually even the next line only makes sense if we show the IDs (even a subset) in the error message

docs/build/manage-workflows/n8n-packages/import-a-package.md
```diff
@@ -0,0 +1,279 @@
...
+		"sourceId": "f9e1e962bba26edf357042c12f367dab17f4c63ec368e1f39cbf7ce43c8b9f8b",
+		"exportedAt": "2026-08-04T09:14:02.117Z"
+	},
+	"workflows": [
+		{ "sourceWorkflowId": "wf9Zx1", "localId": "wf9Zx1", "name": "Send email",
+			"projectId": "prProd4", "parentFolderId": null,
+			"activeVersionId": "<version-id>", "publishing": { "state": "published" }, "status": "created" },
+		{ "sourceWorkflowId": "wf7Kq2", "localId": "wf7Kq2", "name": "Triage inbound",
+			"projectId": "prProd4", "parentFolderId": null,
+			"activeVersionId": "<version-id>", "publishing": { "state": "published" }, "status": "created" }
```
ireneea, line 199: The default behaviour is "unpublished" right? Should we put that default in these examples?

docs/build/manage-workflows/n8n-packages/export-a-package.md
```diff
@@ -0,0 +1,200 @@
...
+| `n8n.audit.n8n-package.export.failed` | The export failed |
+
+The success event carries the user and the `workflowIds`, `folderIds`, and `projectIds` that actually ended up in the package, so it reflects folder contents and auto-included sub-workflows rather than what you asked for. Each list is present only when it isn't empty.
+
+The failure event carries the user, the IDs you requested, and a `reason` of `access-denied`, `entity-not-found`, `blocked`, or `validation`.
+
+Log streaming events carry IDs but no counts. The telemetry events carry the per-entity counts instead, and no IDs.
+
+{% hint style="info" %}
+n8n emits the success event before it streams the archive to you. If the download then fails part-way, the same request emits the failure event too, so one export can produce both. A client that disconnects before the download finishes emits neither an error nor a failure event.
```
ireneea, line 193: Not sure if I'm understanding this correctly

So when a client triggers the export we fire a success event. Then the download starts and If that fails while the client is still  connected I we fire a failure event. If the download fails when the client is disconnected, we don't fire any event.  

Is that correct?

docs/connect/n8n-cli.md
geemanjs: They get replaced at runtime I believe 👍

ireneea (approved):

docs/build/manage-workflows/n8n-packages/export-a-package.md
geemanjs: Went and checked this one, and I think the docs claim holds, though my wording invited the objection so I have tightened it in #5178.

The IDs are definitely built. `package-export.errors.ts` splits them out:

```ts
const message = `${missingIds.length} ${entityType}(s) not found or not accessible. Export aborted.`;
const description = `Missing ${entityType} IDs: ${displayedIds.join(", ")}...`;
```

But `description` does not reach the client. `classifyHttpError`'s `userError` variant is `{ kind, message }`, with nowhere to put it, and `serializePublicApiError` emits `{ message }` alone. The handler's `throw new UserError(error.message, { description: error.description })` looks like it should surface them, and I suspect that is what we are both remembering, but it is inert one layer later.

The existing integration test pins the behaviour: `n8n-packages-export.test.ts` asserts `deniedResponse.body` equals `notFoundResponse.body` for two requests carrying different IDs, which can only pass because the description is absent from the body.

So the workaround line does still make sense. I have narrowed "aren't surfaced anywhere" to "aren't in the response" and added the message shape so it is concrete. Very happy to be wrong if you know a path I have missed, and if we would rather the IDs came back it is a serializer change rather than a docs one.

docs/build/manage-workflows/n8n-packages/import-a-package.md
geemanjs: Good catch, the example was wrong. The value is a third thing though.

The command above the example passes no `--workflow-publishing-policy`, so `preserve-published-state` applies, and `workflow-publishing-policy.ts` only returns `publish` for `status === "updated" && currentlyPublished && sourcePublished`. A created workflow can never satisfy that, so it is always `noop`, which maps to `state: "unchanged"` with a null `activeVersionId`.

Not `unpublished`, interestingly: that is only reported after an actual `deactivateWorkflow`, which can happen to a workflow we updated or archived, never to one we just created.

Fixed in #5178, along with a sentence saying why and pointing at `publish-all` / `match-source` for getting new workflows published.

docs/build/manage-workflows/n8n-packages/export-a-package.md
geemanjs: Your model is right in shape, and the docs were wrong in the middle. Sorry, that hint needed better checking.

`streamPackageExport` rejects on exactly one thing, an error on the source archive stream:

```ts
stream.on("error", reject);
res.on("finish", () => resolve(res));
res.on("close", () => {
	if (!res.writableFinished) stream.destroy();
	resolve(res);
});
stream.pipe(res);
```

So a disconnect, a reset, or a proxy timeout all land on `close` and resolve. No failure event, no error. The only way one request emits both is a server-side error while building or compressing the archive, which classifies as `validation`.

To answer directly: success fires as soon as the archive is built, before the first byte. A broken transfer fires nothing further, whether or not the client is still connected. The both-events case is a server-side archive error, not a network one.

The bit worth documenting, which I have now added: the audit trail records a successful export for a download the user never received. Rewritten in #5178.

PR #5136, opened 2026-07-31
Document the Simplified Custom Auth credential for the HTTP Request node

r00gm:
## Summary

Documents the new **Simplified Custom Auth** generic credential type added in [n8n#35056](https://github.com/n8n-io/n8n/pull/35056).

Changes:

* `docs/integrations/builtin/credentials/httprequest.md`: adds Simplified Custom Auth to the supported authentication methods list, and adds a "Using Simplified Custom Auth" section following the structure of the existing generic auth sections (intro, field list, worked example).
* `docs/integrations/builtin/core-nodes/n8n-nodes-base.httprequest/README.md`: adds the new type to the generic credentials list.

## Notes for review

* Hold until [n8n#35056](https://github.com/n8n-io/n8n/pull/35056) is merged.
* The availability hint says n8n 2.34.0, the next minor cut from master. Adjust if the feature lands in a different release.
* Vale reports no errors on both files; the remaining warnings are pre-existing.

---
## Summary by cubic
Documents the new Simplified Custom Auth credential for the HTTP Request node. Adds an AI Assistant–led creation flow, a guided setup, and examples; available in n8n 2.35.0.

- **New Features**
  - Added “Using Simplified Custom Auth,” led by the AI Assistant and aligned with the credential modal: template with `{{placeholders}}`, field settings (Label, Secret/Plain, Required, Hint), Test URL, accepted status codes, Documentation URL (used by the AI Assistant), API host scoping, header example, required/optional placeholder behavior, and redaction after save.
  - Added Simplified Custom Auth to the supported authentication methods in both the credential doc and the HTTP Request node README.

<sup>Written for commit fb918f7529b54eee315d81d34eaa1c60c9a227ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-docs/pull/5136?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #5182, opened 2026-08-07
fix: replace dead OAuth1 authorization-flow link (DOC-2203)

its-imad:
Fixes [DOC-2203](https://linear.app/n8n/issue/DOC-2203).

## Problem

The `link-check` (Lychee) CI job fails on a dead external link in the HTTP Request credentials doc:

`docs/integrations/builtin/credentials/httprequest.md:101` linked to `https://oauth1.wp-api.org/docs/basics/Auth-Flow.html`, which returns **401**. The whole `oauth1.wp-api.org` domain is gated and returns 401 even with a browser User-Agent, so the resource is gone for readers too, not just CI bots.

This is pre-existing (surfaced while reviewing #5136), and fails `link-check` on every PR that touches this file.

## Fix

Point the "OAuth1 authorization flow" link at [RFC 5849 §2](https://datatracker.ietf.org/doc/html/rfc5849#section-2), the canonical spec for OAuth1's redirection-based authorization flow (verified 200). The adjacent `oauth.net/1/` link is unchanged.

## Test

- `curl` on the new link returns 200.
- No content/style changes beyond the single URL.

### DIFF
--- a/docs/integrations/builtin/credentials/httprequest.md
+++ b/docs/integrations/builtin/credentials/httprequest.md
@@ -98,7 +98,7 @@ To configure this credential, enter:
 
 For most OAuth1 integrations, you'll need to configure an app, service, or integration to generate the values for most of these fields. Use the **OAuth Redirect URL** in n8n as the redirect URL or redirect URI for such a service.
 
-Read more about [OAuth1](https://oauth.net/1/) and the [OAuth1 authorization flow](https://oauth1.wp-api.org/docs/basics/Auth-Flow.html).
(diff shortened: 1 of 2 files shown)


its-imad:
Closing without merging — the original link resolves again.

`https://oauth1.wp-api.org/docs/basics/Auth-Flow.html` now returns **200** (verified with and without a User-Agent). The earlier failure was a transient outage on that site (the whole `oauth1.wp-api.org` domain was 401ing), not a permanently dead resource. Since the existing link works, this change isn't needed. If the site flaps again we can revisit swapping it for [RFC 5849 §2](https://datatracker.ietf.org/doc/html/rfc5849#section-2).

PR #5130, opened 2026-07-29
docs: document Azure Key Vault cloud environment and custom endpoint settings

MarcL:
## Summary

Documents the new **Azure Cloud** selector and custom endpoint fields added to the Azure Key Vault external secrets provider in [n8n-io/n8n#35214](https://github.com/n8n-io/n8n/pull/35214).

The provider previously hardcoded the commercial-cloud vault suffix (`vault.azure.net`) and Entra authority (`login.microsoftonline.com`), so it couldn't reach sovereign-cloud vaults (for example Azure US Government). The docs now cover:

- The **Azure Cloud** dropdown: Azure Public Cloud (default), Azure US Government, Azure China, and Custom, with a table mapping each to its vault URL suffix and Microsoft Entra authority host.
- The **Custom** option: a full **Vault URL** field plus an optional **Authority Host** override, for setups such as Azure Stack or proxied environments.
- A note that existing configurations are unaffected (they default to Azure Public Cloud).

Only the Azure Key Vault section of `docs/administer/manage-credentials/use-external-secret-stores.md` changes.

## Follow-up before merge

- The availability hint uses a `<version>` placeholder in two spots. Fill these in once n8n-io/n8n#35214 merges and the release version is known.

## Related

- Docs ticket: LIGO-901
- Feature ticket: LIGO-791
- Feature PR: https://github.com/n8n-io/n8n/pull/35214

## Preview

https://docs.n8n.io/administer/~/revisions/jOUthA8uXdaGGjMys19L/manage-credentials/use-external-secret-stores#azure-key-vault

CleanShot 2026-07-29 at 18 30 36

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Documents the Azure Cloud selector and custom endpoint fields for Azure Key Vault in `n8n` 2.35.0. Adds a cloud table with vault suffixes and full https Entra authority hosts; use Vault Name for Azure Public/US Government/China, or provide a full Vault URL and optional Authority Host in Custom; earlier versions connect to Public only and existing configs stay on Public (LIGO-791, LIGO-901).

<sup>Written for commit 6b875b49cb8e15940daafcb2867fb9593ad262be. Summary will update on new commits.</sup>

[...]

PR #5206, opened 2026-08-12
docs(changelog): one entry per thing, restructure 2.19 and below

Johannesishere:
Phase 2 of the changelog cleanup started in #5198, covering 2.19 and below, plus two page-wide formatting passes.

## Part 1: one entry per thing, 2.19 and below

Same three rules as #5198: one entry per thing, `###` only for items that make no sense outside their parent, and qualification decided separately from length.

**Five invented umbrellas split.** Their H2 was a name that existed only to hold two or more equal-weight features, with no lead prose because there was nothing to lead. No leads were written for them; they were split into their real constituent entries.

| Umbrella | Became |
| --- | --- |
| 2.19 IdP role mapping and instance bootstrapping | IdP role mapping inside n8n + Instance bootstrapping |
| 2.16 Embedded access and execution data redaction | Token exchange authentication for embedded access + Execution data redaction |
| 2.8 Personal space policies and finer-grained governance | Personal space policies + Inspect a role's permissions before assigning it |
| 2.4 TLS for Syslog log streaming and credential updates via API | TLS support for Syslog log streaming + Update credentials via API |
| 2.2 Finer-grained workflow permissions and richer audit events | More granular workflow permissions in custom project roles + Log streaming: more audit events |

**One roundup merged.** Project-scoped external secrets shipped in halves: instance-admin setup in 2.11, full team access in 2.13. The later completes the earlier, so it is one entry with per-version attribution inside it, dated to the newer half.

**Six items dropped** to Release notes, which already lists every one of them: Public API improvements (2.16), Folder-based filtering in the push and pull dialog (2.13), Custom roles Assignments tab and Workflow execute as a separate permission scope (2.11), and stronger external secrets validation, improved API auditability, and more granular workflow permissions (2.8).

Entries at 2.19 and below go from 15 to 23, and the 21 orphan H3s become 13 entries. The count rises rather than falls because the five umbrellas were concealing ten separately-shareable features behind names nobody shipped. The volume of *things documented* does fall: 33 items become 25.

Every dropped item was re-tested from scratch against "would we have written this up if it were the only thing in the release?", including at short-entry length, which did not exist when an earlier pass looked at this material. What went is an API mirror of an existing UI action, a filter added to a dialog, a tab inside an already-documented feature, security hardening rather than a capability, and the thinnest of three scope splits.

### Availability, verified against the docs

Three entries needed hints the old umbrella structure had lost or never carried:

- Both log streaming entries (2.4 Syslog TLS, 2.2 audit events) now carry `**Availability:** Enterprise.`, per [Log streaming](https://github.com/n8n-io/n8n-docs/blob/main/docs/administer/observe-and-log/stream-logs-to-external-systems.md): "Log Streaming is available on all Enterprise plans."
- The restored custom roles entry carries the same, per [Create custom roles](https://github.com/n8n-io/n8n-docs/blob/main/docs/administer/manage-users-and-access/set-permissions-and-roles-rbac/create-custom-roles.md).

The 2.8 and 2.11 shared hints, which enumerated features that no longer live together, were rewritten to match what remains. The Syslog TLS entry also gained the real setup step from the destination reference: set the transport protocol to TLS rather than the default UDP, and supply the PEM CA certificate.

### One date fix

`release-notes.md` gives 2026-03-03 for n8n 2.11. The new roundup inherited 2026-03-02 from the pre-existing "Easier credential setup on Cloud" entry, which had the same wrong date. Both now match. A full audit of every date on the page against Release notes found one more, untouched here and left alone as out of scope: **Time Saved node** says 2025-12-16 where Release notes gives 2025-12-15.

## Part 2: drop the `***` dividers

All 47 removed. The H2 headline plus the bold `**Released:**` line under it already marks an entry boundary; the horizontal rules added weight without adding separation.

## Part 3: bold only where it earns its place

The repo style guide says "UI elements: bold" (`docs/contribute/style-guide-for-n8n-docs.md`, Text formatting). The page was both over- and under-applying it, and using italics as a competing convention for the same job. Bold now marks four things only:

1. The structural `**Released:**` and `**Availability:**` labels.
2. A UI control in an imperative step: "set **Authentication** to **Microsoft OAuth2 (Graph)**", "click **Compare changes**".
3. A settings navigation path: `**Settings > External Secrets**`.
4. The lead-in term of a bullet in a list of parallel items.

Everything else is plain, including node, resource, credential, and option names in a sentence that merely reports what shipped. What went: bold as emphasis ("in **preview**"), bold on brand names (Moonshot Kimi, Alibaba Cloud Model Studio), bold as a fake sub-heading ("What you can do:", "How to use it:", "How to:", "How to connect:"), and 15 descriptive UI mentions. Italic UI labels in the 2.5 Chat node entry became bold, so UI labels are formatted one way on the page rather than two. Also normalized `Settings → SSO` to the `>` separator used everywhere else.

Bold spans went from 172 to 122: 64 structural, 18 bullet labels, 40 UI controls in instructions.

## Verified

- 47 entries, each with a well-formed `**Released:**` line and exactly one blank line as separation.
- Every `#n8nXYZ` anchor resolves against `docs/changelog/release-notes.md`. `#n8n283` does not exist, so the two 2.8.3 entries point at `#n8n28`, the closest valid anchor.
- Page stays version-sorted newest first, not date-sorted.
- Only 3 `###` subsections left, all legitimate: MiniMax and NVIDIA Nemotron under New model providers, Oracle vector store under the Oracle entry.
- Every referenced `.gitbook/assets/` image exists, including `custom_roles_selector.png`, in use again.
- No em dashes, no absolute `docs.n8n.io` links, no "beta", no duplicate headlines. New prose measures median 16 words per sentence, max 26.
- All checks green, including `runner / vale`, which I expected to fail structurally on a multi-entry changelog PR and did not.

The `changelog-entry` skill has been updated for the divider removal and the bold rule. Note that Fabian's pipeline carries its own standalone prompt outside this repo, so its PRs will keep arriving with dividers and over-applied bold until that prompt is edited.

### DIFF
--- a/docs/changelog/README.md
+++ b/docs/changelog/README.md
@@ -612,7 +612,7 @@ Workflow settings updates are also logged with the specific parameters that chan
 
 ## Time Saved node
 
-**Released:** 2025-12-16 in [n8n 2.1](release-notes.md#n8n21)
+**Released:** 2025-12-15 in [n8n 2.1](release-notes.md#n8n21)
 
 Previously, teams could only track a single fixed time saved value for each workflow regardless of which path an execution takes. The new Time Saved node enables more precise time savings calculations where different execution paths save different amounts of time.
 
--- a/docs/changelog/README.md
+++ b/docs/changelog/README.md
@@ -581,6 +581,20 @@ This release builds on recent updates to the permissions model, including [custo
 
 ***
 
+## Inspect a role's permissions before assigning it
+
+**Released:** 2026-02-13 in [n8n 2.8.3](release-notes.md#n8n28)
+
+The project role selector now splits built-in system roles and [custom roles](https://app.gitbook.com/s/wMJrGrimpx3PxCJpUswm/manage-users-and-access/set-permissions-and-roles-rbac/create-custom-roles) into separate sections, so a long list of custom roles no longer buries the defaults. Hovering over any role shows a summary of its configured permissions, with an option to open the full permission details. Previously you had to leave the assignment flow and open the role itself to check what you were about to grant.
+
+<figure><img src=".gitbook/assets/custom_roles_selector.png" alt="System roles and custom roles are now displayed in separate sections"><figcaption><p>System roles and custom roles are now displayed in separate sections</p></figcaption></figure>
+
+{% hint style="info" %}
+**Availability:** Enterprise.
+{% endhint %}
+
+***
+
 ## Human-in-the-loop for AI tool calls
 
 **Released:** 2026-01-26 in [n8n 2.6](release-notes.md#n8n26)
@@ -640,7 +654,11 @@ Human in the loop for the Chat node
 
 **Released:** 2026-01-12 in [n8n 2.4](release-notes.md#n8n24)
 
-The Syslog log streaming destination now supports TLS over TCP. Previously it sent log events unencrypted, which ruled it out for SIEM and observability platforms that require encrypted transport. Enable TLS on the destination and point it at your platform's TLS port, and n8n streams audit and execution events over an encrypted connection.
+The Syslog [log streaming](https://app.gitbook.com/s/wMJrGrimpx3PxCJpUswm/observe-and-log/stream-logs-to-external-systems) destination now supports TLS over TCP. Previously it sent log events over plain UDP or TCP, which ruled it out for SIEM and observability platforms that require encrypted transport. Set the destination's transport protocol to TLS instead of the default UDP, and supply the PEM-formatted CA certificate for the connection.
+
+{% hint style="info" %}
+**Availability:** Enterprise.
+{% endhint %}
 
 ***
 
@@ -679,6 +697,10 @@ Log streaming now includes additional audit events to improve visibility into op
 This update adds events for manual workflow cancellations and workflow activation/deactivation (publish/unpublish), variable lifecycle events (create/update/delete), and user management actions (including enabling/disabling 2FA).
 
 Workflow settings updates are also logged with the specific parameters that changed (for example, selecting a new error workflow), instead of a generic "updated" event.
+
+{% hint style="info" %}
+**Availability:** Enterprise.
+{% endhint %}
 
 ***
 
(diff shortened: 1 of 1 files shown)


Johannesishere:
Valid, thanks. `release-notes.md` gives **2026-03-03** for n8n 2.11, so the roundup attribution was wrong.

Fixed, and the same wrong date turned out to be in the pre-existing "Easier credential setup on Cloud" entry, which is where the roundup inherited it from. Fixing only the roundup would have left two different dates for n8n 2.11 on the same page, so both now read 2026-03-03.

Auditing every date on the page against `release-notes.md` turned up one more, untouched by this PR: **Time Saved node** says `2025-12-16 in [n8n 2.1]`, but Release notes gives 2025-12-15. Left alone as out of scope here.

Johannesishere:
Merged `main` in (just #5130, a different file, no conflict).

On the P3 bold finding: the observation is accurate but it is measured against the repo-wide rule, and this page deliberately narrows it. The changelog rule, agreed for this PR, is context-based rather than label-class-based:

- **Bold** when the reader is being told to act on the control (imperative step), when it is a settings navigation path, or when it is the lead-in term of a bullet.
- **Plain** when the label appears in a sentence that merely reports what shipped.

That is why the pairs cubic lists differ. `**Response Mode**` and `**Using Response Nodes**` sit in "add the Response Mode option and set it to Using Response Nodes", an instruction. `Only run if`, `After Decision`, `Get Diff` and `Get Patch` sit in "the node gains an expression-based Only run if option" and "Get Diff and Get Patch fetch the raw diff", which describe rather than instruct. Same class of label, different job in the sentence.

`Send a message and wait for response` is the clearest case: bold as a bullet lead-in (L547), bold in the numbered step where you pick it (L557), plain in the "Keep in mind" note that only refers to it (L559). All three follow the rule.

I checked all 34 UI-label mentions on the page against this rule and found no violations, so the page is internally consistent. The repo style guide ("UI elements: bold", `docs/contribute/style-guide-for-n8n-docs.md`) stays the rule for procedural docs; the changelog is narrative, and bolding every literal label is what made entries read speckled, which is the problem this pass set out to fix. The narrower rule is now written into the `changelog-entry` skill and into the drafting prompt so it holds for future entries.

No change from this one.

PR #5234, opened 2026-08-18
Document custom OAuth scopes for Typeform credentials

TMWilds:
## Summary

Documents the new "Custom Scopes" toggle on the Typeform OAuth2 credential (added in [n8n-io/n8n#36486](https://github.com/n8n-io/n8n/pull/36486)), which lets users request Typeform API permissions beyond the three scopes n8n requests by default. Also fixes a pre-existing duplicate step number (two steps numbered "10.") in the same OAuth2 setup list.

## Related

https://linear.app/n8n/issue/CE-2321


BGZStephen (approved):
LGTM. Default scopes and the toggle names match the credential.

PR #5264, opened 2026-08-21
fix(linking): add guidance on internal linking

RoRoJ:
## Summary

Adds explicit rules for cross-linking pages within a section to the style guide and the `n8n-docs-author` skill, then applies them to two sections (**Flow logic** and **Understand workflows**) as worked examples for reviewers to check against.

## Rules added

- **"In this section"**: every section landing page lists all its direct children as a bullet list (link + one-line description).
- **"Related pages"**: every child page gets a footer section linking back to its parent and sideways to each direct sibling, as bare links (no descriptions — the reader has already read a page from the cluster, so the link text alone carries enough signal). Authors can add other genuinely related pages here at their discretion, even outside the current section.
- **Exceptions**: flat reference collections (one page per data type, etc.) skip sibling links entirely; sections with more than 8-10 siblings drop the sibling list and link back to the parent only, since a full list past that size is mostly boilerplate and an O(n²) maintenance burden.

## Applied to two sections to show what it looks like

- **`docs/build/flow-logic/`**: 8 pages get "Related pages" footers linking back to the `Flow logic` overview and each other.
- **`docs/build/understand-workflows/`**: a nested section with two subsections (`understand-executions/`, `workflow-components/`). Applying the rules here also surfaced and fixed incomplete "in this section" style lists on all three READMEs (some pages weren't listed at all, e.g. `find-your-workflow-id.md` was missing from `workflow-components`).

## Test plan

- [ ] Spot-check a few pages in both sections render correctly on the GitBook preview
- [ ] Confirm relative links (`./`, `../`) resolve correctly, especially in the nested `understand-workflows` subsections

### DIFF
--- a/docs/build/flow-logic/README.md
+++ b/docs/build/flow-logic/README.md
@@ -15,6 +15,17 @@ layout:
 
 n8n allows you to represent complex logic in your workflows.
 
+## In this section <a href="#in-this-section" id="in-this-section"></a>
+
+* [Split with conditionals](split-with-conditionals.md): route items down different branches with the IF and Switch nodes.
+* [Merge data](merge-data.md): combine multiple data streams back into one.
+* [Loop](loop.md): repeat an action until a condition is met, or until all items are processed.
+* [Wait](wait.md): pause a workflow's execution and resume it later.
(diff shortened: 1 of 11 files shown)


its-imad (approved):
lgtm

PR #5274, opened 2026-08-24
docs: update compression node config var defaults

Flexicon:
Docs side of the V3 breaking change in default config var values as implemented here: https://github.com/n8n-io/n8n/pull/36935/changes

---
## Summary by cubic
Updates the environment variable docs to the v3 defaults for the Compression node (CAT-4212). Old defaults were 2 GiB and 5000 entries; new defaults are 256 MiB and 1000 entries to reduce decompression risk.

- Migration: To keep pre-v3 behavior, set `N8N_COMPRESSION_NODE_MAX_DECOMPRESSED_SIZE_BYTES=2147483648` and `N8N_COMPRESSION_NODE_MAX_ZIP_ENTRIES=5000`.

<sup>Written for commit 38ebf5c74cca1986243a7a102ccabed8a23e2963. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/n8n-io/n8n-docs/pull/5274?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


RoRoJ (commented):
This looks good to me, do you think it's worth adding to the v3 breaking changes page [here](https://docs.n8n.io/changelog/v30-breaking-changes#security) as a fourth bullet point?

- **Lower Compression node decompression limits.** Default `N8N_COMPRESSION_NODE_MAX_DECOMPRESSED_SIZE_BYTES` drops from 2 GiB to 256 MiB, and default `N8N_COMPRESSION_NODE_MAX_ZIP_ENTRIES` drops from 5,000 to 1,000.
  - **What to do:** If your workflows decompress archives larger than 256 MiB or with more than 1,000 entries, set these variables explicitly to their previous values (`2147483648` and `5000`) before upgrading to n8n 3.0.

Flexicon:
Seems like a good idea to me 👌🏻  I think @HumanistSerif mentioned the docs team will add this to some breaking changes page and I was to just assign the PR to him for now - I guess this is the page he meant?

**fyi:** the default values I've updated here will only be live once v3 drops.

HumanistSerif (approved):

PR #5297, opened 2026-08-27
docs: add exception to internal link check

RoRoJ:
## Summary

Fixes a false positive in the `internal-links` CI check for the `./` link pattern.

`[Configure n8n](./)` is a documented way to link to a folder's own parent page (its `README.md`) — see the style guide's "Link to the current page's parent page" section. GitBook resolves this bare directory-self-reference natively, but the checker's Rule 1 (internal links must end in `.md`) had no exception for it, so it was flagged as `no-md-extension` even though the link works correctly on the live site.

Also adds clarification in style guide that not all parent pages are directly reachable via ./

## Changes

- `.github/scripts/check_internal_links.py`: added a special case that resolves a bare `./` target against the linking file's own directory `README.md`, before the `.md`-extension rule is applied. Falls through to the existing `missing-target` / anchor checks so a genuinely broken `./` (missing `README.md`) is still caught.
- Updated the module docstring's Rule 1 description to note this documented exception.
- - Updated style guide

## Test plan

- [x] Ran the exact failing CI command (`check_internal_links.py` against the changed `docs/deploy/...` files) — now exits 0 with no errors.
- [x] Ran the checker against the full `docs/` tree — only 2 pre-existing, unrelated errors remain (`absolute-path` in a webhook doc, `cross-space-broken` in the license doc), confirming no regressions.

### DIFF
--- a/docs/contribute/style-guide-for-n8n-docs.md
+++ b/docs/contribute/style-guide-for-n8n-docs.md
@@ -545,7 +545,7 @@ Use the file name on its own:
 
 **Link to the current page's parent page**
 
-Use `./` only if the current folder's landing page is a `README.md` — it points there:
+Use `./` only if the current folder's landing page is a `README.md`, it points there:
 
 ```
 [link to a parent page](./)
--- a/docs/contribute/style-guide-for-n8n-docs.md
+++ b/docs/contribute/style-guide-for-n8n-docs.md
@@ -146,7 +146,7 @@ Retrieved on its own, a section that leans on its neighbours arrives stripped of
 Connect each page to the others on its topic. Explicit, descriptive links let an agent follow a path directly instead of guessing a URL, and they group your pages into a topic cluster that AI search reads as a signal of depth.
 
 * **Always link the prerequisites and the next step**, at minimum.
-* **Link parents and children both ways.** An overview or section landing page lists and links to every child page; each child links back to its parent with `./`.
+* **Link parents and children both ways.** An overview or section landing page lists and links to every child page; each child links back to its parent.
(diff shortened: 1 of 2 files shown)


its-imad (approved):
LGTM Small suggestion either way: when ./ fails but a sibling <foldername>.md exists, the error could say "this folder's parent page isn't a README.md, link ../<foldername>.md directly" instead of target not found: ./ -> .../README.md. Happy to do that in a follow-up.

I've also updated the two Notion pages (internal link checker CI, and the internal linking guide) to document the ./ exception.

docs/contribute/style-guide-for-n8n-docs.md
```diff
@@ -545,12 +545,14 @@ Use the file name on its own:
 
 **Link to the current page's parent page**
 
-Use `./`, which points at the parent page — the `README.md` landing page of the current folder (`understand-workflows`):
+Use `./` only if the current folder's landing page is a `README.md` — it points there:
```
RoRoJ:
```suggestion
Use `./` only if the current folder's landing page is a `README.md`, it points there:
```

RoRoJ:
> LGTM Small suggestion either way: when ./ fails but a sibling .md exists, the error could say "this folder's parent page isn't a README.md, link ../.md directly" instead of target not found: ./ -> .../README.md. Happy to do that in a follow-up.
> 
> I've also updated the two Notion pages (internal link checker CI, and the internal linking guide) to document the ./ exception.

Yes, this did occur to me - its' a bit messy that some child pages don't have README.md parents, but I think we'll have to work with that for now. Happy for you to follow that up next time you edit.

PR #5318, opened 2026-09-03
Update latest and next version numbers

its-imad:
Automated stable/beta version-snippet update (npm dist-tags -> GitBook reusable content).

This PR is auto-merged once all required checks pass. A human is pinged only if something looks off: npm moved unexpectedly, or the PR changes more than the two version numbers.

PR #5320, opened 2026-09-03
fix(docops): report auto-merges without repo labels (branch+marker + comment dedupe)

its-imad:
## What

Follow-up fix from the **first live auto-merge** ([#5318](https://github.com/n8n-io/n8n-docs/pull/5318), merged in ~3 min, approved by the `n8n-assistant` App 🎉).

That run exposed a gap: the merged-sweep filtered merged PRs by a `docops:auto-merge` label that **doesn't exist** in this repo, so `gh pr edit --add-label` silently no-op'd and the sweep never matched the PR — the **outcome webhook** (Slack success + Supabase log) never fired. The merge itself was fine; only the async *reporting* half was silently skipped.

## Fix — no repo labels at all

We deliberately don't add DocOps-specific labels (keeps the repo's label set clean):

- `arm_auto_merge` no longer adds a label.
- `announce_merged` now detects merged snippet PRs by the **fixed title + branch regex + body marker** (all already present) and **dedupes via a hidden PR comment** (`<!-- docops:outcome-sent -->`) instead of a label.
- Escalation/close still reuse the **existing** `status:do-not-merge` / `status:outdated`.

CI/automation only, no docs content. Tests unchanged and green (16 verdict + 20 helper).

> #5318's missed report was already replayed manually to the webhook, so its Slack/Supabase entries exist and Linear DOC-2272 is Done (Linear's native GitHub integration also closed it on merge).

Ref: DOC-2269

### DIFF
--- a/.github/scripts/docops_auto_merge.py
+++ b/.github/scripts/docops_auto_merge.py
@@ -346,11 +346,13 @@ def announce_merged(repo, cfg):
     fires (Slack success + Supabase; Linear usually already closed via its own GitHub
     integration). Detection = fixed title + branch + body marker; dedupe = a hidden PR
     comment. No repo labels are used, created, or required."""
-    # The search is already narrowed to snippet PRs by their fixed title, so this window
-    # spans months of them at ~1/day; a merge is reported within one sweep (15 min) anyway.
+    # The search is already narrowed to snippet PRs by their fixed title (~1/day), and
+    # `gh pr list --limit` paginates internally up to the cap. 500 keeps well over a year
+    # of them retryable, so an undelivered outcome cannot age out of the window in
+    # practice (that would need the webhook down for hundreds of snippet PRs).
     merged = gh_json(["pr", "list", "-R", repo, "--state", "merged",
                       "--search", "in:title Update latest and next version numbers",
-                      "--limit", "100",
+                      "--limit", "500",
                       "--json", "number,headRefName,body,mergeCommit"]) or []
     for p in merged:
         if not cfg["branch_re"].match(p.get("headRefName", "")):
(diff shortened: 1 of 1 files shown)


its-imad:
all 4 were valid, all 4 addressed in `6e8dd80`.

**2 (re-reporting #5318) — the real one.** Correct: #5318's outcome was replayed manually via the webhook, which left no `OUTCOME_MARKER`, so the first sweep would have double-reported it. I seeded the marker comment on #5318 directly ([comment](https://github.com/n8n-io/n8n-docs/pull/5318#issuecomment-5525884284)) and verified it's present, so the sweep will skip it.

**3 (marker written even when the POST fails).** `post_outcome` now returns delivery status; the dedupe marker is only written when the POST actually landed. A webhook failure now leaves no marker and retries on the next sweep. (It returns `True` when no webhook is configured — nothing to deliver, nothing to retry.)

**4 (POST succeeds, marker write fails → repost).** Fully solving this needs receiver-side idempotency, which is a change on the n8n side. For now the failure is no longer silent: we log a visible `WARNING` naming the PR. The residual worst case is one duplicate Slack ping (no data loss, no wrong merge), which I judged proportionate here rather than adding a Supabase read to the webhook. Noted as a follow-up to make the webhook idempotent per PR.

**1 (capped search).** Raised 15 → 100. Worth noting the search is already narrowed to snippet PRs by their fixed title, and a merge is reported within one sweep (15 min), so falling out of the window needed 15 snippet PRs merging inside one interval. Still, the wider cap is free.

Tests unchanged and green (16 verdict + 20 helper).

its-imad:
Addressed in `f8a17e6`: widened the merged-sweep cap **100 → 500**.

Worth noting for context: `gh pr list --limit` paginates internally up to the cap (it is not a single page), and the search is already narrowed to snippet PRs by their fixed title (~1/day), while the sweep runs every 15 min and reports everything it finds. So an undelivered outcome ageing out now requires the webhook to be down for hundreds of snippet PRs — well over a year — which I documented in the comment rather than adding unbounded pagination.

PR #5321, opened 2026-09-03
docs: Add a self-hosted example for the n8n domain placeholder

its-imad:
Fixes [DOC-2258](https://linear.app/n8n/issue/DOC-2258/mcp-client-examples-your-n8n-domain-only-ever-shows-a-cloud-example).

## What

On [MCP client connection examples](https://docs.n8n.io/connect/connect-to-n8n-mcp-server/mcp-client-examples), every client section ends with a "Here, replace:" list defining `<your-n8n-domain>`. That definition appeared **10 times** and every one gave only a Cloud example, even though instance-level MCP is available on Cloud and self-hosted (the enablement section is headed "For Cloud and self-hosted instances"). A self-hosted reader got an example that looks nothing like their instance.

Each definition now names both hosting types with a concrete example.

## Also fixed

The examples included the `https://` scheme, but the code samples already write `https://<your-n8n-domain>/mcp-server/http`. Substituted literally that gave `https://https://your-instance.app.n8n.cloud/mcp-server/http`. The nine `<your-n8n-domain>` examples are now bare hosts. The Google ADK bullet keeps the scheme, since `N8N_INSTANCE_URL` is interpolated as a full base URL.

## New wording

> `<your-n8n-domain>`: Your n8n domain, without the `https://` prefix. On n8n Cloud this is `your-instance.app.n8n.cloud`. When self-hosted, it's the domain that serves your n8n editor, for example `n8n.example.com`. You can find your full server URL in n8n under **Settings** > **Instance-level MCP > Connect a client > Server URL**.

The ticket suggested wording with an em dash; the [style guide](https://docs.n8n.io/contribute/style-guide-for-n8n-docs) bans them, so this splits into sentences instead.

The ticket also floated factoring the sentence into a reusable-content include. Skipped: `docs/reusable-content/` holds only a stub, and GitBook include targets are authored in the GitBook UI rather than in git.

## Out of scope

- `N8N_INSTANCE_URL = "https://localhost:5678"` in the Google ADK Python sample doesn't work as written (local n8n serves HTTP). Split out as [DOC-2273](https://linear.app/n8n/issue/DOC-2273/google-adk-mcp-sample-sets-n8n-instance-url-to-httpslocalhost5678); this PR leaves the code block alone.
- The landing-page cards rendering `https:///mcp-server/http` are DOC-2254.

## Checks

- Local Vale on the file: 0 warnings on the changed lines (one pre-existing `Vale.Spelling` error on "workspace's" at line 227, untouched by this PR).
- No page moves or renames, so no `SUMMARY.md` or redirect changes.


its-imad:
Addressed the cubic P3, and updated the branch off `main`.

The finding is valid: the bullets say to drop the `https://` prefix while every snippet hardcodes it, so a plain-HTTP instance substitutes into an unusable URL. Verified that the case is real, since the server URL comes from `urlService.getInstanceBaseUrl()` and follows `N8N_PROTOCOL` / `WEBHOOK_URL` / `N8N_EDITOR_BASE_URL`, and n8n allows HTTP for localhost (see the localhost exemption in `update-allowed-redirect-uris.dto.ts`).

I did not take the suggested fix of changing the prefix in each snippet, since that means editing nine code blocks and turns a wording change into a restructure. Instead there is one sentence in the page intro:

> The examples below all use `https://`. If your instance serves plain HTTP, such as a local install at `http://localhost:5678`, use `http://` instead. Copying the full **Server URL** from n8n gives you the right prefix either way.

Worth noting this was never a regression. The previous text produced `https://https://your-instance.app.n8n.cloud/...` for everyone.

Vale on the file: 0 warnings, plus the same pre-existing `Vale.Spelling` error on "workspace's" that this PR does not touch.

HumanistSerif (approved):

PR #5329, opened 2026-09-04
fix: add missing blank lines before headings glued to preceding blocks

bennycode:
Headings (such as `##` and `###`) immediately following a GitBook `{% endhint %} ` tag with no blank line in between can fail to render (see screenshot).

## Example

image

The "Git overview" heading is broken on the live docs page.

## Failing Renderings

- [Light versioning](https://docs.n8n.io/connect/create-nodes/build-your-node/reference/versioning)
- [How do I set up logging in n8n?](https://docs.n8n.io/deploy/host-n8n/keep-n8n-running/set-up-logging)
- [Turn LDAP off](https://docs.n8n.io/administer/manage-users-and-access/verify-user-identity/connect-ldap)
- [Git overview](https://docs.n8n.io/administer/use-source-control-and-environments/use-git-in-n8n)
- [Permissions](https://docs.n8n.io/build/manage-workflows/share-with-others)

PR #5330, opened 2026-09-04
ci: add markdownlint check for headings glued to preceding blocks

bennycode:
Adds a CI check for the bug fixed in #5329: a heading immediately following `{% endhint %}` with no blank line in between, which can fail to render on the live site.

## What this adds

- `.github/markdownlint-rules/heading-blank-line-after-endhint.js`: a custom markdownlint rule. It walks the micromark token stream and flags any real heading (`atxHeading` token) whose preceding raw line is exactly `{% endhint %}`.
- `.markdownlint-cli2.jsonc`: enables only this rule.
- `.github/workflows/markdownlint.yml`: runs it on PRs via `markdownlint-cli2-action`.

## Why this shape

- **Scoped to `{% endhint %}` only, not a general "blank line above heading" rule.** Other GitBook block tags (`{% hint %}`, `{% step %}`, `{% column %}`) have their own "first line is a title" behavior and render a glued heading correctly as a styled title, not as broken text — confirmed live for the `{% step %}`/`{% column %}` cases. Only `{% endhint %}` has been confirmed to actually break rendering.
- **Uses the micromark token stream instead of a raw line/regex scan.** An `atxHeading` token only exists for a heading markdownlint's own parser recognizes as real, so a heading written inside a fenced code example (e.g. to illustrate this exact bug, the kind of thing the style guide already does for other GitBook syntax) is excluded automatically, with no need to hand-track fence delimiters.
- **No path-based exclusions needed anywhere.** Runs clean across all of `docs/` as-is.

Draft until #5329 merges — this branch is stacked on top of it so the check passes cleanly against the fixed content.


its-imad (changes requested):
Great catch on this. 
thank you for both spotting it and immediately building something to stop it coming back. That's the right instinct!

One thought on where the check lives. All our docs checks are Python scripts in .github/scripts/ with tests in tests/, running on the files changed in the PR (internal-links.yml is the closest example). We don't have Node in CI at all yet, so I'd rather not add a whole toolchain for one rule if we can avoid it.

I'd also suggest we give it a slightly wider home, maybe check_gitbook_syntax.py, with your {% endhint %} rule as the first one in it. GitBook formatting that silently breaks a page probably isn't a one-off, so the next one we find would just be a few lines added rather than another workflow.


Concretely, here's roughly what I'd do:

- Add .github/scripts/check_gitbook_syntax.py, with your {% endhint %} rule as its first rule
- Add .github/scripts/tests/test_check_gitbook_syntax.py with a passing case, a failing case, and one inside a code fence
- Register the test job in .github/workflows/script-tests.yml
- Replace markdownlint.yml with a workflow modelled on internal-links.yml: changed files only, job named runner / gitbook-syntax, actions pinned to a SHA like our other workflows
- Drop .markdownlint-cli2.jsonc and the custom JS rule

Happy to do the port myself and let you review it, if that's easier. Either way #5329 first, as you said.

bennycode:
Hi @its-imad and thank you for your valuable feedback!

I was also hesitant to add Node.js to the pipeline which is why I didn't add any npm scripts directly but were purely relying on a GitHub Action (`DavidAnson/markdownlint-cli2-action`) to do the job. But I agree that a Python rule would fit more what is already present.  

I merged #5329. Happy for you to do the toolchain port to Python and to review it once it's up. 👍

its-imad:
Ported this to Python in #5359 (same protection, no Node toolchain, run on changed files like `internal-links.yml` per our discussion). Full credit to you for the bug and the original check, noted on the new PR. Closing this in favor of #5359 - would appreciate your review there.

PR #5332, opened 2026-09-04
ci(release-notes): open the weekly PR as the n8n-assistant App

Johannesishere:
Moves authorship of the weekly release-notes PR off a personal credential and onto the `n8n-assistant` App.

### Why

The PR is generated by **Publish release notes to docs** on the internal n8n instance, which opened it using my own GitHub credential. That makes me both the author and the only person reliably looking at it — and `main` requires one approving review, which GitHub won't let an author give themselves. In practice the PR could only land through an admin bypass, and that bypass is gone.

With the App as author, an ordinary human review satisfies the ruleset. Nothing about the review itself changes; it just starts counting.

### How

n8n keeps generating and committing the files exactly as today, then fires a `repository_dispatch` carrying the branch, title and body. This workflow opens the PR as the App using that title and body verbatim, so the generated entry counts and warnings survive unchanged.

### Verified before proposing

Ran it from a throwaway branch first: the App token minted with `permission-pull-requests: write`, and #5331 came out authored by `app/n8n-assistant`. The probe branch and its workflow file are already deleted; nothing from it reached `main`.

### Security posture

Deliberately mirrors `docops-auto-merge.yml`:

- App token, not `github-actions[bot]`; actions SHA-pinned; `permissions: {}` on the job.
- `repository_dispatch` only ever executes the default-branch copy of the workflow, so a branch cannot run its own edited version.
- `client_payload` is caller-controlled, so it is passed through `env` and never interpolated into the script body, and `branch` is pattern-checked against `^release-notes/[0-9]{8}-[0-9]{4}$` — a dispatch cannot open a PR from an arbitrary ref.
- Idempotent: a retried dispatch reuses the open PR rather than opening a second one.
- It opens a PR and stops. No approve, no merge, no auto-merge.

### Notes for reviewers

@its-imad — this is a second consumer of the App alongside your auto-merge POC, and it lands near your `# phase 2 may add a dedicated author bot` note. Worth a look for whether you'd rather these converge on one identity. Note the direction is the opposite of yours: yours has the App *approve*, this has it *author*, so the human review still happens.

@Matsuuu — you introduced the App in #5283; flagging since this is a new usage of it.

I'll flip the n8n side to dispatch once this is on `main`. Until then the current behaviour is unchanged.


Matsuuu (approved):

its-imad:
Looks good, no changes needed from my side.

On converging identities: I'd keep them separate. If `n8n-assistant` ever becomes the *author* of the version-snippet PRs, its approval stops counting and auto-merge breaks, same self-approval rule you're working around here. So the phase 2 author bot in my note should be a different identity, and `n8n-assistant` stays the approver. Your usage and mine don't overlap.

They also can't collide today. My workflow needs all three of the body marker, the `^doc-[0-9]+-update-version-snippet$` branch name, and an allowed author. Your PRs match none of them, so auto-merge will never pick one up. That's also why cubic's P1 is low impact here: a forged dispatch can only leave an unreviewed PR open, it can't reach the merge path.

PR #5335, opened 2026-09-05
docs(http-request): document multipart file name and 415 binary body issues

fstandhartinger:
## What this changes

Adds two sections to the HTTP Request node's common issues page, covering the two failures people hit when they send a file:

1. **`Your uploaded file arrives with the wrong file name`** — with **Body Content Type** set to **Form-Data**, the `filename` in the multipart part comes from the binary data, not from the **Name** field (which sets the form field name). Nothing in the current docs says where that file name comes from, so when an API keys uploads by file name the failure looks unexplainable.
2. **`Unsupported Media Type (415) when you send a file`** — **Body Content Type** > **n8n Binary File** sends the raw file as the entire body with the binary item's MIME type and no multipart envelope. APIs that require `multipart/form-data` reject it with a 415, which reads like a bug in the multipart support rather than a different body type.

Neither the [HTTP Request node page](https://docs.n8n.io/integrations/builtin/core-nodes/n8n-nodes-base.httprequest) nor its common issues page mentions either behaviour today.

## Why

This is a recurring stumbling block for anyone posting a file to an API that identifies uploads by file name, for example HTML-to-PDF services that require the file to be called `index.html`. The failure gives no hint at the cause: the node succeeds in building the request, and the API returns a generic 400 "file is required" or a 415.

## How this was verified

Measured on a fresh **n8n 2.37.7** from npm (SQLite, workflows created and executed over `POST /rest/workflows/:id/run`) against a local recorder that writes the exact bytes of each request, on September 5, 2026. Four workflows, identical apart from the one setting under test:

| Body Content Type | File name set | What went on the wire |
| --- | --- | --- |
| Form-Data | none (**Convert to Text File** default) | `Content-Type: multipart/form-data`, part `name="files"`, `filename="file.txt"` |
| Form-Data | **File Name** = `index.html` | same, `filename="index.html"` — accepted, 16,466-byte PDF returned |
| Form-Data | set in a **Code** node (`item.binary.data.fileName`) | `filename="index.html"` |
| n8n Binary File | `index.html` | `Content-Type: text/plain`, 199 bytes, no boundary, no parts |

That matches the source: `filename: binaryData.fileName ?? 'file'` and `contentType: binaryData.mimeType` in the `formBinaryData` branch of `HttpRequestV3.node.ts`, and the `binaryData` branch setting the body to the raw buffer with `'content-type': itemBinaryData.mimeType ?? 'application/octet-stream'`.

## Checklist

- Vale passes on the added text. The one remaining error on this file (line 36, `could not`) is pre-existing and sits inside a heading that quotes the product's error message verbatim, so I left it alone.
- Both internal links resolve (`../n8n-nodes-base.converttofile.md`, `../n8n-nodes-base.code/README.md`).
- No new page, so no `SUMMARY.md` change needed.
- Style guide followed: sentence-case headings, second person, contractions, no anchor markup on the new headings.

## Disclosure

I maintain a community node for PDF generation, which is how I ran into both behaviours. There's no mention or link to it here, and neither section names any product. An AI assistant helped with the measurements and the draft, on my account and with my consent.

<a href="https://cubic.dev/pr/n8n-io/n8n-docs/pull/5335?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


fstandhartinger:
Unblocked from my side and ready for review:

- **CLA signed** — the check was failing because I ran `/cla-check` before actually signing. Signed now, and the bot reports `CLA Check passed`.
- **Branch updated** onto `main` (it was behind).
- **Section anchors added.** Every other heading on this page carries the inline `<a href="#slug" id="slug"></a>`; my two new sections were missing it, so they had no anchor target. Fixed in 59da028.

All checks green, and I checked the two sections in the GitBook preview.

Disclosure, since it's relevant to the content: I build an HTML/Markdown-to-PDF API and the n8n community node that calls it, which is how I ran into both of these. Neither section links to or mentions it — they're about the HTTP Request node's own behaviour, which I reproduced in n8n before writing them up.

RoRoJ (approved):

RoRoJ:
Thanks for your contribution, looks great.

PR #5343, opened 2026-09-07
docs(changelog): error workflow executions no longer count towards your quota (n8n 2.38)

lucasnilssonn8n:
Adds a changelog entry for the 2.38 change that excludes error workflow runs from the execution quota, on every plan.

- Placed at the top as the newest entry (Released 2026-09-01, the date on the Release notes page; anchor `release-notes.md#n8n2381` checked, since 2.38 appears under its first patch).
- Links to the Handle errors gracefully page via its GitBook space URL.
- Follows the format in the PMM "How to changelog" Notion page: H2 headline, Released line, no availability hint since it applies to every plan.
- Supersedes #5342.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5374, opened 2026-09-10
docs(changelog): Business and Enterprise got the error workflow quota change in 2.28

lucasnilssonn8n:
Follow-up to #5343. The last paragraph of the error workflow quota entry now says the change has been live on Cloud since 2.38, and on self-hosted applies to Business and Enterprise instances from 2.28.0 (1.123.60 on v1). The unconfirmed "all other instances from 2.38.0" clause is gone.

### DIFF
--- a/docs/changelog/README.md
+++ b/docs/changelog/README.md
@@ -50,7 +50,7 @@ Until now each of those runs counted like any other execution. Handling a failur
 
 Attach an error workflow to a workflow in its **Workflow Settings** and its runs are excluded from the count. One error workflow can serve as many workflows as you like.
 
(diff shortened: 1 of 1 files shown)


its-imad (approved):
lgtm 

Only thing left is cubic's P2: the entry opens with "on every plan" but then limits self-hosted to Business and Enterprise. Cubic's suggested wording fixes it -> "on every Cloud plan, and on self-hosted Business and Enterprise." Apply that and merge whenever.

lucasnilssonn8n:
@its-imad It's on all plans but community doesn't have billable executions (because it's free) so it doesn't make a difference for them.

its-imad:
got it thanks

PR #5398, opened 2026-09-15
CAT-65: docs: describe actual isEmpty() behavior for booleans and numbers

Flexicon:
## Summary

Resolves the documentation part of [CAT-65](https://linear.app/n8n/issue/CAT-65/expressions-isempty-is-incorrect-when-called-on-false-values). The Catalysts team decided to keep the current `isEmpty()` runtime behavior and update the docs to match it instead of shipping a breaking change in v3.

Number and Boolean have no type-specific `isEmpty()` or `isNotEmpty()` in `packages/workflow/src/extensions`, so calls fall through to the generic versions, which treat any falsy value as empty. Verified against current `master` and confirmed on a production instance:

| Input | `isEmpty()` | `isNotEmpty()` |
|---|---|---|
| `false`, `0`, `NaN` | `true` | `false` |
| `true`, `1` | `false` | `true` |
| `""`, `[]`, `{}`, `null`, `undefined` | `true` | `false` |
| DateTime | `false` | `true` |

## Changes

- **boolean.md, number.md**: fix the `isEmpty()` descriptions and the `false` / `0` examples, and add a warning hint that `isEmpty()` isn't a null check, with the direct-comparison and **If** node **exists** alternatives. Add the missing `isNotEmpty()` sections.
- **string.md, array.md, object.md**: mention `undefined` alongside `null`.
- **datetime.md**: add the `isEmpty()` and `isNotEmpty()` sections that exist in the source but were missing from the docs.
- **README.md**: update the index descriptions to match, and add index entries for the four new sections.


docs/build/work-with-data/transform-data/expression-reference/boolean.md
```diff
@@ -12,7 +12,11 @@ layout:
 
 ## _`Boolean`_.**`isEmpty()`** <a href="#booleanisempty" id="booleanisempty"></a>
 
-**Description:** Returns <code>false</code> for all booleans. Returns <code>true</code> for <code>null</code>.
+**Description:** Returns <code>true</code> if the boolean is <code>false</code>, <code>null</code>, or <code>undefined</code>. Returns <code>false</code> if the boolean is <code>true</code>.
+
+{% hint style="warning" %}
+`isEmpty()` isn't a null check. On a boolean it treats `false` as empty, so `{{ $json.flag.isEmpty() }}` returns `true` for both a missing field and a field set to `false`. To test only for a missing value, compare directly, for example `{{ $json.flag === null }}`, or use the **exists** operator in the **If** node.
```
vterzic, line 18: > To test only for a missing value, compare directly, for example `{{ $json.flag === null }}`

Should this be `{{ $json.flag === unedfied }}` or `{{ $json.flag == null }}`

vterzic (approved):
it looks good. 

just one comment/question

docs/build/work-with-data/transform-data/expression-reference/boolean.md
Flexicon: Yeah, one can use either I guess but I decided to just leave it on the side of not filling the box with too much text 🙂  We can add both if you think it reads better.

vterzic: I guess we can keep it as is. I think that the message is properly conveyed at the end of the day.

PR #5399, opened 2026-09-15
docs: Add a System tasks page with the Prometheus metrics they expose

looorent:
## Summary

Documents the Prometheus metrics for system tasks.

### New page: `docs/deploy/host-n8n/configure-n8n/system-tasks.md`

- What system tasks are and how they run: in-memory timers on the leader vs. the durable scheduler, told apart by the `mode` label
- `N8N_METRICS_INCLUDE_SYSTEM_TASK_METRICS` and the label set (`task`, `mode`, `result`, `reason`), including the no-instance-label aggregation rule
- All ten `n8n_system_task_*` series, grouped into run health, schedule health, and in-memory scheduling
- Nine PromQL queries to start from (dead task, late vs. its own cadence, never-succeeded, failure rate, p95 duration, hung run, skips by reason, lost schedule, p99 lag)

## Related Linear tickets, Github issues, and Community forum posts

https://linear.app/n8n/issue/CAT-4178


looorent:
Converted to draft until the grafana link lands.

looorent:
@cubic-dev-ai  re-execute your review

looorent:
Nothing left to act on from cubic's three open threads; resolving them.

- **Parent/child links (P3)** — fixed in a3cb3fc: `configure-n8n/README.md` now has content-refs for both `system-tasks.md` and `durable-scheduler.md`, `docs/deploy/SUMMARY.md` has the nav entry, and both pages end with `See [Configure n8n](./)`.
- **Dashboard link 404 (both P2s)** — no longer a 404: the dashboard is published at `n8n-io/n8n-observability` under `dashboards/grafana/n8n-system-tasks/` on `main`. This PR was held as a draft until that landed, which is why the link pointed at nothing when cubic first checked.

vterzic (approved):
👍

docs/deploy/host-n8n/configure-n8n/system-tasks.md
```diff
@@ -0,0 +1,90 @@
...
+
+- **In-memory mode**, `in_memory`: the task runs from a timer in the leader main instance's process. This is the default mode. Only the leader runs these timers: they stop when an instance steps down and start again on the instance that takes over. A run whose time passes while the instance is down doesn't happen. If the process runs but sleeps, a suspended container for example, the timer fires once for all the occurrences it slept through instead of replaying them one by one.
+- **Durable mode**, `durable`: the task runs from the [durable scheduler](durable-scheduler.md)'s database-backed queue, so any main instance can pick up the run and a restart doesn't drop it. Set both `N8N_SCHEDULER_ENABLED` and `N8N_SCHEDULER_SYSTEM_TASKS_ENABLED` to `true` to use this mode, and give every instance in your setup the same values. Only tasks that support durable mode move over; the rest stay on their in-memory timers. As of n8n 2.41.0, no system task supports durable mode, so every task runs from an in-memory timer.
```
RoRoJ, line 20: When comparing technical accuracy against the published codebase, AI review came back with:

**False as a blanket statement**. Two system tasks already declare `durable = true` in the current source: `instance-registry-stale-member-cleanup (packages/cli/src/modules/instance-registry/stale-member-cleanup.task.ts:24) `and `jti-cleanup(packages/cli/src/modules/token-exchange/services/jti-cleanup.task.ts:23)`. 

It's true that the three tasks the page actually names as examples elsewhere (`execution-pruning-soft-delete`,` insights-compaction`, `license-renewal`) are all still `durable = false`— so the claim is accurate only for that specific family, not for "no system task." 

**Suggestion**: Rescope the sentence to the named maintenance-job family: 

As of n8n 2.41.0, the pruning, compaction, and license-renewal tasks all still run in-memory

— or drop the absolute claim entirely.

looorent: @RoRoJ Yes, that's true. But we did it deliberately, because around 20 tasks are going to become "durable" over the next few weeks. So for this first version of the documentation we decided to assume that "nothing is durable", and to change that once the 20 tasks have been migrated, rather than having to go into too much detail now.

Wdyt?

RoRoJ: That makes sense, OK for me

RoRoJ (approved):

PR #5417, opened 2026-09-18
ci: check the docs.n8n.io links hardcoded in the n8n product repo

its-imad:
Closes part of [DOC-2318](https://linear.app/n8n/issue/DOC-2318/rebuild-the-links-from-n8n-to-docs-checker-resolve-path-plus-anchor-on). This is the collection half; the reporting half is a pair of n8n workflows on docs-prod.

## Problem

Node notices, credential hints and error messages in `n8n-io/n8n` link to docs.n8n.io pages and anchors. Nothing checks them:

- `lychee.yml` / `lychee-pr-check.yml` only scan `docs/`, and `lychee.toml` excludes `^https://docs\.n8n\.io/.*`.
- `internal-links.yml` does check anchors, but only on markdown changed in an n8n-docs PR.
- Neither runs against the `n8n` repo.

So when a heading is renamed, the product link dies silently. Benny hit this in n8n-io/n8n#38992; DOC-2317 was a second case the same day.

## What this adds

- `.github/workflows/product-links.yml` - weekly (Mon 07:48 UTC, 30 min after the lychee sweep) plus `workflow_dispatch`. Checks out this repo and `n8n-io/n8n` (depth 1, sparse `packages`).
- `.github/scripts/product_links_report.py` - extracts every docs.n8n.io URL from the n8n source and resolves each one against this repo's markdown.
- `.github/scripts/tests/test_product_links_report.py` - 31 network-free unit tests, wired into `script-tests.yml`.

Every URL lands in one bucket:

| Kind | Meaning |
|---|---|
| `ok` | path resolves to a docs file, and the anchor (if any) matches |
| `broken-anchor` | page exists, no heading carries that id - **actionable now** |
| `empty-anchor` | link ends in a bare `#` |
| `unresolved-path` | no file backs the path: either dead or alive only via a GitBook redirect. n8n live-checks these and splits them |
| `templated` | URL built at runtime (`${...}`), not checkable |
| `generated` | GitBook-generated subtree (OpenAPI API reference) with no markdown source |

## Why resolution is offline

Anchors are the point of this ticket, and an HTTP status check cannot see them. Resolving against the markdown makes the anchor check exact. It reuses `check_internal_links.heading_ids`, so this checker and the in-repo one agree on slugs by construction rather than by convention - including its policy of staying quiet when a heading has no explicit `id=` or an `{% include %}` can't be resolved, so a gap costs coverage instead of a false alarm.

The one thing offline resolution can't do is tell a genuinely dead path from one a redirect still catches. That split needs a live request, so n8n does it.

## Trial run

Run against my local checkout, which only has `packages/cli` and `editor-ui` (no `nodes-base`): 2453 files, 63 unique URLs, and it found the exact bug class the ticket describes:

> `pyodide-removed.rule.ts` tells users to see `.../n8n-nodes-base.code/#python-native-beta` - the heading id is `python-native`. Live in production today.

It also correctly skipped `https://docs.n8n.io/not-in-registry/`, which looks dead but is a fixture in `n8n-docs.tool.test.ts`. That is why `SKIP_DIRS` exists.

51 of the 63 were `unresolved-path`, nearly all pre-GitBook-migration paths still riding a redirect (`/code/variables/`, `/data/data-mapping/...`, `/2-0-breaking-changes/`). The full repo including `nodes-base` will be considerably larger.

## Before the schedule is useful

The `DOCOPS_PRODUCT_LINKS_WEBHOOK_URL` repo secret has to be added. Without it the job goes red with a clear error rather than passing quietly - same contract as `lychee_report.py`. `WEBHOOK_USER` / `WEBHOOK_PASSWORD` already exist.

No docs content changes, so no GitBook preview to review.


RoRoJ (approved):
There's a Cubic comment which I'll leave you to look at, but otherwise LGTM

